PR #1, opened 2021-02-03
some 4554 support

hobokenn:
image

### DIFF
--- a/src/hooks/engine.rs
+++ b/src/hooks/engine.rs
@@ -28,6 +28,8 @@ pub static build_number: Pointer<unsafe extern "C" fn() -> c_int> = Pointer::emp
     Patterns(&[
         // 6153
         pattern!(55 8B EC 83 EC 08 A1 ?? ?? ?? ?? 56 33 F6 85 C0),
+        // 4554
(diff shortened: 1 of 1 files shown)


src/hooks/engine.rs
```diff
@@ -154,7 +168,7 @@ pub static Host_InitializeGameDLL: Pointer<unsafe extern "C" fn()> = Pointer::em
     // Alternatively, find LoadEntityDLLs() and go to the parent function.
     Patterns(&[
         // 6153
-        pattern!(E8 ?? ?? ?? ?? 8B 0D ?? ?? ?? ?? 33 C0),
+        pattern!(E8 ?? ?? ?? ?? 8B 0D ?? ?? ?? ?? 33 C0 83 F9 01),
```
YaLTeR, line 171: Was it non-unique in 4554?

hobokenn: yes, it was

YaLTeR: Interesting, I wonder what the other function was.

YaLTeR:
Thanks, this is awesome!

src/hooks/engine.rs
hobokenn: one was `SV_WriteClientdataToMessage` and the other one was `VideoMode_SwitchMode`, this is what it matched for the latter one 
image

YaLTeR: I see, okay.

PR #3, opened 2021-02-21
add bxt_shake_remove

hobokenn:


src/hooks/engine.rs
```diff
@@ -307,6 +307,20 @@ pub static Sys_VID_FlipScreen: Pointer<unsafe extern "C" fn()> = Pointer::empty_
     ]),
     my_Sys_VID_FlipScreen as _,
 );
+pub static V_ApplyShake: Pointer<unsafe extern "C" fn(*mut c_float, *mut c_float, c_float)> =
+    Pointer::empty_patterns(
+        b"V_ApplyShake\0",
+        // To find, search for "ScreenShake". This is ClientDLL_Init(), near the bottom there are
+        // two similar function calls, one is using our string as the 1st param and the 2nd param as
+        // another function, open it. This is V_ScreenShake(), right above it is V_ApplyShake().
```
YaLTeR, line 317: > right above it is `V_ApplyShake()`

This isn't a good search method because the compiler can reorder functions in the binary, so this might not be the case for some other DLL version. However, I don't see a better search method for now, so let's leave it.

In the future I was planning to find the `cl_enginefuncs` somewhere and get function pointers from that (their order doesn't change for most of the engine versions), then this pattern can be replaced.

YaLTeR (commented):
Thanks! Which of the screen shakes does this remove?

hobokenn:
> 
> 
> Thanks! Which of the screen shakes does this remove?

like the env_shake ones i guess, the punchangle stuff stays the same

src/hooks/engine.rs
hobokenn: > This isn't a good search method because the compiler can reorder functions in the binary
i did check some older dlls and it was always there

PR #19, opened 2021-08-18
Added bxt_disable_skybox

### LINKED ISSUE #16: Skybox overrides view with bxt_wallhack 1
https://user-images.githubusercontent.com/58108407/127754721-72457173-3901-4848-98ea-9661c9103bf1.mp4

SmileyAG:
Closes #16

### DIFF
--- a/src/modules/disable_skybox.rs
+++ b/src/modules/disable_skybox.rs
@@ -1,0 +1,50 @@
+//! `bxt_disable_skybox`
+
+use super::Module;
+use crate::gl;
+use crate::hooks::engine;
+use crate::modules::cvars::{self, CVar};
+use crate::utils::*;
+
+pub struct DisableSkybox;
+impl Module for DisableSkybox {
+    fn name(&self) -> &'static str {
+        "bxt_disable_skybox"
+    }
+
+    fn cvars(&self) -> &'static [&'static CVar] {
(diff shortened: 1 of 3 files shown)


src/modules/disable_skybox.rs
```diff
@@ -0,0 +1,48 @@
...
+        "bxt_disable_skybox"
+    }
+
+    fn cvars(&self) -> &'static [&'static CVar] {
+        static CVARS: &[&CVar] = &[&BXT_DISABLE_SKYBOX];
+        CVARS
+    }
+
+    fn is_enabled(&self, marker: MainThreadMarker) -> bool {
+        engine::R_DrawSkyBox.is_set(marker) && cvars::CVars.is_enabled(marker)
```
YaLTeR, line 21: Also needs a `gl::GL.borrow(marker).is_some()`.

YaLTeR (commented):
I'd investigate whether the issue can be fixed without disabling skybox (probably by moving skybox drawing earlier so it's drawn "below" everything else), but until then this workaround can stay.

src/modules/disable_skybox.rs
SmileyAG: Done.

YaLTeR:
Thanks

PR #28, opened 2021-11-24
engine: Add patterns for 1600

SmileyAG:
Functions which not available to find:

- gEntityInterface
- _Z18Sys_VID_FlipScreenv
- VideoMode_IsWindowed
- VideoMode_GetCurrentVideoMode

@hobokenn, I provided link to that build in Discord DM for help complete this bxt-rs support.


src/hooks/engine.rs
```diff
@@ -142,7 +142,7 @@ pub static Cmd_AddMallocCommand: Pointer<
         // 4554
         pattern!(56 57 8B 7C 24 ?? 57 E8 ?? ?? ?? ?? 8A 08),
         // 1600
-        pattern!(56 57 8B 7C 24 ?? 57 E8 ?? ?? ?? ?? 8A 08 83 C4 04 84 C9 74 ?? 57 68 ?? ?? ?? ?? E8 ?? ?? ?? ?? 83 C4 08 5F 5E C3),
+        pattern!(56 57 8B 7C 24 ?? 57 E8 ?? ?? ?? ?? 8A 08 83 C4 04 84 C9 74 ?? 57 68 ?? ?? ?? ?? E8 ?? ?? ?? ?? 83 C4 08 5F ?? C3),
```
YaLTeR, line 145: Yeah, this is an unfortunate issue with the macro. For some reason the compiler doesn't just pass the contents as is, but tries to parse them as numbers, and things like `5E` get recognized as an incomplete number in exponential notation (e.g. 5E10 or 5E-3). I'm not sure why this happens, but until it's somehow fixed in bxt-macros we'll have to live with masking these bytes.

YaLTeR:
Thanks

PR #31, opened 2021-12-07
engine: extend 4554 pattern for Cmd_AddMallocCommand

SmileyAG:
Turns out that the pattern from the 4554 engine matches for two functions on the 1600 build, and I couldn't immediately notice it.


YaLTeR:
Thanks

PR #37, opened 2022-02-01
force_fov: don't allow to use values below 1 to prevent wallhack state

SmileyAG:
https://github.com/YaLTeR/BunnymodXT/pull/254


YaLTeR:
Thanks

PR #40, opened 2022-05-08
Added bxt_disable_loading_text

SmileyAG:
Related to https://github.com/YaLTeR/BunnymodXT/pull/289

### DIFF
--- a/src/hooks/engine.rs
+++ b/src/hooks/engine.rs
@@ -529,6 +529,17 @@ pub static S_TransferStereo16: Pointer<unsafe extern "C" fn(c_int)> = Pointer::e
     ]),
     my_S_TransferStereo16 as _,
 );
+pub static SCR_DrawLoading: Pointer<unsafe extern "C" fn()> = Pointer::empty_patterns(
+    b"SCR_DrawLoading\0",
+    // To find, search for "cz_worldmap" string in Steampipe DLL.
+    // This is SCR_DrawPause. Right below that function is SCR_DrawLoading.
+    // This pattern also works for the pre-Steampipe builds.
+    Patterns(&[
+        // 6153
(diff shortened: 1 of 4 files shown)


src/hooks/engine.rs
```diff
@@ -529,6 +529,17 @@ pub static S_TransferStereo16: Pointer<unsafe extern "C" fn(c_int)> = Pointer::e
     ]),
     my_S_TransferStereo16 as _,
 );
+pub static SCR_DrawLoading: Pointer<unsafe extern "C" fn()> = Pointer::empty_patterns(
+    b"SCR_DrawLoading\0",
+    // To find, search for "cz_worldmap" string in Steampipe DLL.
+    // This is SCR_DrawPause. Right below that function is SCR_DrawLoading.
```
YaLTeR, line 535: Is there a way without relying on function order in the DLL? It's not guaranteed to be constant.

SmileyAG: That order is constant between steampipe DLLs at the moment.

YaLTeR: Still is there a different way to find the function without relying on the order?

SmileyAG: Well, I got some idea of how it can be found at other way, but it looks kinda more complicated:

```
Search "transition" string, there is two functions with that string: SCR_BeginLoadingPlaque and SCR_EndLoadingPlaque.

You should go to SCR_EndLoadingPlaque, it can be recognized by having much less code than in SCR_BeginLoadingPlaque.

Find second variable inside of that function, it would be scr_drawloading boolean.

Now to references of variable and find other function with shortest code in it, that would be SCR_DrawLoading function. 
```

YaLTeR: Could you add this to the comment below? Like, "Another way to find this would be as follows."

SmileyAG: Done.

YaLTeR:
Thanks

PR #45, opened 2022-06-23
Shrink unsafe blocks

peamaeq:
In this function you use the unsafe keyword for almost the entrie function body. 

We need to mark unsafe operations more precisely using unsafe keyword. Keeping unsafe blocks small can bring many benefits. For example, when mistakes happen, we can locate any errors related to memory safety  within an unsafe block. This is the balance between Safe and Unsafe Rust. The separation is designed to make using Safe Rust as ergonomic as possible, but requires extra effort and care when writing Unsafe Rust. 

Hope this PR can help you.
Best regards.
**References**
https://doc.rust-lang.org/nomicon/safe-unsafe-meaning.html 
https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html

### DIFF
--- a/src/modules/capture/mod.rs
+++ b/src/modules/capture/mod.rs
@@ -184,8 +184,8 @@ static BXT_CAP_STOP: Command = Command::new(
 );
 
 fn cap_stop(marker: MainThreadMarker) {
+    let mut state = STATE.borrow_mut(marker);
     unsafe {
-        let mut state = STATE.borrow_mut(marker);
         if let State::Recording(ref mut recorder) = *state {
             match recorder.record_last_frame() {
                 Ok(()) => {
--- a/src/modules/demo_playback.rs
+++ b/src/modules/demo_playback.rs
@@ -146,10 +146,11 @@ fn find_demos(prefix: PathBuf) -> Result<Vec<(PathBuf, usize)>, String> {
 pub fn set_next_demo(marker: MainThreadMarker) {
     let mut demos = DEMOS.borrow_mut(marker);
 
-    unsafe {
+    
         // Safety: no engine functions are called while the reference is active.
-        let cls_demos = &mut *engine::cls_demos.get(marker);
+        let cls_demos = unsafe { &mut *engine::cls_demos.get(marker) };
 
+    
         match demos.pop() {
             Some(demo) => {
                 // Replace the first startdemos entry with the next demo and set the next demo as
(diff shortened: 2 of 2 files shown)


src/modules/capture/mod.rs
```diff
@@ -184,8 +184,8 @@ static BXT_CAP_STOP: Command = Command::new(
 );
 
 fn cap_stop(marker: MainThreadMarker) {
+    let mut state = STATE.borrow_mut(marker);
     unsafe {
-        let mut state = STATE.borrow_mut(marker);
```
YaLTeR, line 188: I am sometimes using unsafe blocks as regular blocks that happen to need unsafe, i.e. to also restrict the variable scope. In particular, here `state` is dropped at the end of the `unsafe {}` block, which is important, as right below `STATE` is accessed again, which will now cause a `RefCell` non-exclusive-mut-borrow panic.

Therefore, I'm afraid, this change is incorrect.

src/modules/demo_playback.rs
```diff
@@ -146,10 +146,11 @@ fn find_demos(prefix: PathBuf) -> Result<Vec<(PathBuf, usize)>, String> {
 pub fn set_next_demo(marker: MainThreadMarker) {
     let mut demos = DEMOS.borrow_mut(marker);
 
-    unsafe {
+    
         // Safety: no engine functions are called while the reference is active.
-        let cls_demos = &mut *engine::cls_demos.get(marker);
+        let cls_demos = unsafe { &mut *engine::cls_demos.get(marker) };
 
+
```
YaLTeR, line 153: You seem to have added extra blank lines and forgot to run `cargo fmt`.

YaLTeR (changes requested):
> Keeping unsafe blocks small can bring many benefits. For example, when mistakes happen, we can locate any errors related to memory safety within an unsafe block.

This is not always the case. The scope of unsafe is an abstraction boundary. A mistake leading to memory safety can lie in a completely safe function. Ralf Jung describes this very well in [The Scope of Unsafe](https://www.ralfj.de/blog/2016/01/09/the-scope-of-unsafe.html).

In bxt-rs, given it's such an unusual use-case for Rust, the unsafe boundary has a bit of a loose definition with unsafe being necessary in a lot of places. Therefore it doesn't always make sense to strive to contain the unsafe blocks to the smallest possible subexpression. I do generally try to do that, but within reason.

YaLTeR:
Please feel free to fix the changes and reopen.

PR #47, opened 2022-09-03
capture: Use wrapping multiplication for sample pairs

hobokenn:
It was overflowing because of the bugged gauss charge sound in the Steam version.


YaLTeR:
Thanks! Guess I gotta do a 3.1 with this.

PR #50, opened 2022-10-06
CS1.6 related changes for TAS Optimizer

khanghugo:
Hmmm, it doesn't look up to standard from my standard.

### DIFF
--- a/bxt-strafe/src/lib.rs
+++ b/bxt-strafe/src/lib.rs
@@ -51,7 +51,7 @@ pub struct Player {
     pub in_duck_animation: bool,
     /// Ducking animation timer.
     pub duck_time: i32,
-    // Stamina time, for game with stamina
+    /// For game with stamina like CS1.6
     pub stamina_time: f32,
 }
 
@@ -269,7 +269,6 @@ mod tests {
             use_slow: true,
             has_stamina: false,
             ducktap_slow: false,
-            
         }
     }
 
--- a/bxt-strafe/src/lib.rs
+++ b/bxt-strafe/src/lib.rs
@@ -51,6 +51,8 @@ pub struct Player {
     pub in_duck_animation: bool,
     /// Ducking animation timer.
     pub duck_time: i32,
+    // Stamina time, for game with stamina
+    pub stamina_time: f32,
 }
 
 impl Player {
@@ -81,6 +83,11 @@ pub struct Parameters {
     pub step_size: f32,
     pub bounce: f32,
     pub bhop_cap: bool,
+    pub bhop_cap_multiplier: f32,
+    pub bhop_cap_maxspeed_scale: f32,
+    pub use_slow: bool,
+    pub has_stamina: bool,
+    pub ducktap_slow: bool,
 }
 
 /// The type of player's position in the world.
@@ -257,6 +264,12 @@ mod tests {
             step_size: 18.,
             bounce: 1.,
             bhop_cap: false,
+            bhop_cap_multiplier: 0.65,
+            bhop_cap_maxspeed_scale: 1.7,
+            use_slow: true,
+            has_stamina: false,
+            ducktap_slow: false,
+            
         }
     }
 
@@ -268,6 +281,7 @@ mod tests {
             ducking: false,
             in_duck_animation: false,
             duck_time: 0,
+            stamina_time: 0.0,
         }
     }
 
@@ -431,8 +445,9 @@ mod tests {
             ducking in any::<bool>(),
             in_duck_animation in any::<bool>(),
             duck_time in 0..1000,
+            stamina_time in 0f32..1315.789429,
         ) -> Player {
-            Player { pos, vel, base_vel, ducking, in_duck_animation, duck_time }
+            Player { pos, vel, base_vel, ducking, in_duck_animation, duck_time, stamina_time }
         }
     }
 
--- a/bxt-strafe/src/lib.rs
+++ b/bxt-strafe/src/lib.rs
@@ -51,7 +51,7 @@ pub struct Player {
     pub in_duck_animation: bool,
     /// Ducking animation timer.
     pub duck_time: i32,
-    /// For game with stamina like CS1.6
+    /// Stamina timer for games like CS1.6.
     pub stamina_time: f32,
 }
 
@@ -84,10 +84,10 @@ pub struct Parameters {
     pub bounce: f32,
     pub bhop_cap: bool,
     pub bhop_cap_multiplier: f32,
-    pub bhop_cap_maxspeed_scale: f32,
-    pub use_slow: bool,
+    pub bhop_cap_max_speed_scale: f32,
+    pub use_slow_down: bool,
     pub has_stamina: bool,
-    pub ducktap_slow: bool,
+    pub duck_animation_slow_down: bool,
 }
 
 /// The type of player's position in the world.
@@ -265,10 +265,10 @@ mod tests {
             bounce: 1.,
             bhop_cap: false,
             bhop_cap_multiplier: 0.65,
-            bhop_cap_maxspeed_scale: 1.7,
-            use_slow: true,
+            bhop_cap_max_speed_scale: 1.7,
+            use_slow_down: true,
             has_stamina: false,
-            ducktap_slow: false,
+            duck_animation_slow_down: false,
         }
     }
 
--- a/bxt-strafe/src/steps.rs
+++ b/bxt-strafe/src/steps.rs
@@ -470,6 +470,7 @@ impl<S: Step> Step for Friction<S> {
             // stamina code is before the clamp_velocity
             if parameters.has_stamina {
                 state.player.vel.x *= (100. - (state.player.stamina_time/ 1000.) * 19.) / 100.; 
+                state.player.vel.y *= (100. - (state.player.stamina_time/ 1000.) * 19.) / 100.; 
             }
         }
 
@@ -526,7 +527,7 @@ impl<S: Step> Step for Jump<S> {
     ) -> (State, Input) {
         if parameters.has_stamina { 
             state.player.stamina_time = 
-                (state.player.stamina_time - (parameters.frame_time * 100.) as f32).max(0.);
+                (state.player.stamina_time - (parameters.frame_time * 1000.) as f32).max(0.);
         }
 
         if input.jump && !state.prev_frame_input.jump && state.place == Place::Ground {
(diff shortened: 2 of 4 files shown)


khanghugo:
https://github.com/YaLTeR/bxt-rs/blob/576e01238156bec1fccaf879b84ad5d6efd0600e/bxt-strafe/src/steps.rs#L730-L732

`state.player.in_duck_animation` condition does not work (as in doing nothing) for CS1.6 duck tap slow but I just want to keep it on parity with BunnymodXT. More or less, it is just a condition that does nothing but only `input.duck` is responsible for that. I am unaware if it does cause any problems. There is an issue with lgagst being weird every 1/1000 bhop block where it refuses to jump with dbg/dbc and keeps on walking in duck unless lgagst is toggled off. I am not sure if this could be the cause. When I encounter that occurrence, I might just try out and see and I hope that is isn't.

khanghugo:
https://github.com/YaLTeR/bxt-rs/blob/a8e93b22cefe5cda17da3f221cefb1a8295d92a1/bxt-strafe/src/steps.rs#L471-L472

Whoops, it is supposed to scale the whole vector.

bxt-strafe/src/lib.rs
```diff
@@ -51,6 +51,8 @@ pub struct Player {
     pub in_duck_animation: bool,
     /// Ducking animation timer.
     pub duck_time: i32,
+    /// For game with stamina like CS1.6
```
YaLTeR, line 54:
```suggestion
    /// Stamina timer for games like CS1.6.
```

bxt-strafe/src/lib.rs
```diff
@@ -81,6 +83,11 @@ pub struct Parameters {
     pub step_size: f32,
     pub bounce: f32,
     pub bhop_cap: bool,
+    pub bhop_cap_multiplier: f32,
+    pub bhop_cap_maxspeed_scale: f32,
```
YaLTeR, line 87:
```suggestion
    pub bhop_cap_max_speed_scale: f32,
```

bxt-strafe/src/lib.rs
```diff
@@ -81,6 +83,11 @@ pub struct Parameters {
...
+    pub use_slow: bool,
```
YaLTeR, line 88:
```suggestion
    pub use_slow_down: bool,
```

bxt-strafe/src/lib.rs
```diff
@@ -81,6 +83,11 @@ pub struct Parameters {
...
+    pub has_stamina: bool,
+    pub ducktap_slow: bool,
```
YaLTeR, line 90:
```suggestion
    pub duck_animation_slow_down: bool,
```

bxt-strafe/src/steps.rs
```diff
@@ -466,6 +466,11 @@ impl<S: Step> Step for Friction<S> {
 
                 state.player.vel *= new_speed / speed;
             }
+
+            if parameters.has_stamina {
+                state.player.vel.x *= (100. - (state.player.stamina_time / 1000.) * 19.) / 100.;
+                state.player.vel.y *= (100. - (state.player.stamina_time / 1000.) * 19.) / 100.;
```
YaLTeR, line 472:
```suggestion
                let factor = (100. - (state.player.stamina_time / 1000.) * 19.) / 100.;
                state.player.vel.x *= factor;
                state.player.vel.y *= factor;
```

bxt-strafe/src/lib.rs
```diff
@@ -81,6 +83,11 @@ pub struct Parameters {
     pub step_size: f32,
     pub bounce: f32,
     pub bhop_cap: bool,
+    pub bhop_cap_multiplier: f32,
+    pub bhop_cap_maxspeed_scale: f32,
+    pub use_slow: bool,
+    pub has_stamina: bool,
```
YaLTeR, line 89: I wonder if maybe we can just get rid of this variable, and set stamina_time = 0 for games without stamina?

src/modules/tas_optimizer/mod.rs
```diff
@@ -331,10 +333,19 @@ fn optim_init(marker: MainThreadMarker, path: PathBuf, first_frame: usize) {
 
     // TODO: get current parameters.
     let parameters = unsafe {
+        let game_dir = CStr::from_ptr(engine::com_gamedir.get(marker).cast())
+            .to_str()
+            .unwrap();
+
+        let is_paranoia: bool = game_dir.eq("paranoia");
+        let is_cstrike: bool = game_dir.eq("cstrike");
+        let is_czero: bool = game_dir.eq("czero");
```
YaLTeR, line 342: We don't want to `com_gamedir.get()` here because it'll panic if `com_gamedir` wasn't found. Instead, in that case we want to assume that there are no game-dir-related effects.
```suggestion
        // Safety: the reference does not outlive this block, and com_gamedir can only be modified
        // at engine start and while setting the HD models or the addon folder.
        let game_dir = engine::com_gamedir
            .get_opt(marker)
            .and_then(|dir| CStr::from_ptr(dir.cast()).to_str().ok())
            .unwrap_or("valve");

        let is_paranoia = game_dir == "paranoia";
        let is_cstrike = game_dir == "cstrike";
        let is_czero = game_dir == "czero";
```

src/modules/tas_optimizer/mod.rs
```diff
@@ -331,10 +333,19 @@ fn optim_init(marker: MainThreadMarker, path: PathBuf, first_frame: usize) {
...
+
+        let pmove: *mut playermove_s = *engine::pmove.get(marker);
```
YaLTeR, line 344: Once again we don't want to depend on `pmove` being found. And let's de-duplicate the `sv_maxspeed` part.
```suggestion
        let max_speed = get_cvar_f32(marker, "sv_maxspeed").unwrap_or(320.);
        let client_max_speed = engine::pmove
            .get_opt(marker)
            .map(|pmove| (**pmove).clientmaxspeed)
            .unwrap_or(if is_paranoia { 100. } else { 0. });
```

src/modules/tas_optimizer/mod.rs
```diff
@@ -350,6 +361,42 @@ fn optim_init(marker: MainThreadMarker, path: PathBuf, first_frame: usize) {
...
+                    (*pmove).clientmaxspeed * get_cvar_f32(marker, "sv_maxspeed").unwrap_or(320.)
+                        / 100.
+                } else if (*pmove).clientmaxspeed > 0.0f32
+                    && get_cvar_f32(marker, "sv_maxspeed").unwrap_or(320.) > (*pmove).clientmaxspeed
+                {
+                    (*pmove).clientmaxspeed
+                } else {
+                    get_cvar_f32(marker, "sv_maxspeed").unwrap_or(320.)
+                }
+            },
```
YaLTeR, line 375:
```suggestion
                if is_paranoia {
                    max_speed * client_max_speed / 100.
                } else if client_max_speed != 0. {
                    max_speed.min(client_max_speed)
                } else {
                    max_speed
                }
            },
```

src/modules/tas_optimizer/mod.rs
```diff
@@ -350,6 +361,42 @@ fn optim_init(marker: MainThreadMarker, path: PathBuf, first_frame: usize) {
...
+            use_slow: !(is_cstrike || is_czero),
+            has_stamina: (is_cstrike || is_czero)
+                && match get_cvar_f32(marker, "bxt_remove_stamina") {
+                    Some(x) => x != 1.0f32,
+                    None => {
+                        con_print(marker, "Cannot find bxt_remove_stamina argument.\n");
+                        false
+                    }
+                },
```
YaLTeR, line 398:
```suggestion
                && get_cvar_f32(marker, "bxt_remove_stamina").map(|x| x != 1.).unwrap_or(true),
```

YaLTeR (changes requested):
Looks good, but needs a few improvements.

khanghugo:
Thanks, this seems reasonable. Let me do that.

bxt-strafe/src/lib.rs
khanghugo: Not sure what you mean by this. There is currently this line responsible for updating stamina.

https://github.com/YaLTeR/bxt-rs/blob/a8e93b22cefe5cda17da3f221cefb1a8295d92a1/src/modules/tas_optimizer/mod.rs#L864

However, it just updates. The variable `has_stamina` makes it meaningful. Otherwise, I guess we have to use the `com_gamedir` to says that stamina only applies to that game? I guess if you want that then sure.

khanghugo:
This has all the changes except the unresolved one, not sure how it would proceed :+1:

bxt-strafe/src/lib.rs
YaLTeR: Yeah, I was thinking put the game dir check here, and set to 0 if it's not cstrike.

khanghugo: Hmmm, from my testing, removing this would be more trouble than it's worth. This would also require checking game in the `bxt-strafe/../step.rs` as well. I attempted to just check whether the stamina is there (stamina is 0) and do actions accordingly but it is possible for CS1.6 to have stamina of 0 too.

khanghugo: This just makes the stamina prediction just weirdly borked.

khanghugo: I guess I have to do actual checking of the game then?

YaLTeR: I thought all stamina operations just do nothing when stamina is 0, requiring no extra checks? Or am I mistaken

khanghugo: Welp, that would complicate things too as it needs thread marker. Would you reconsider this?

YaLTeR: Ah, right, the stamina set on jump. Forgot about that. Yeah let's just leave this as is then

khanghugo: Chat goes so fast that Github doesn't even refresh :+1: maybe it is my adblocker.

src/modules/tas_optimizer/mod.rs
```diff
@@ -333,15 +332,21 @@ fn optim_init(marker: MainThreadMarker, path: PathBuf, first_frame: usize) {
...
+            .and_then(|dir| CStr::from_ptr(dir.cast()).to_str().ok())
+            .unwrap_or("valve");
+        let is_paranoia = game_dir == "paranoia";
+        let is_cstrike = game_dir == "cstrike";
+        let is_czero = game_dir == "czero";
+
+        let max_speed = get_cvar_f32(marker, "sv_maxspeed").unwrap_or(320.);
+        let client_max_speed = engine::pmove
+            .get_opt(marker)
+            .map(|pmove| (**pmove).clientmaxspeed)
```
YaLTeR, line 348: I see BXT uses `GetClientMaxspeed()` for this. I'm not sure using `pmove` is equivalent, isn't it one frame off? Since `pmove` is only filled/updated later in the frame.

khanghugo: I am not aware of it being one frame off. It does sync well on my end. If it is one frame off then it would be one frame off the very first frame of the TAS as weapon speed would be the same the whole run.

khanghugo: Also, it is in the bind, so I just use whatever is there.

YaLTeR:
Thanks!

PR #51, opened 2022-10-27
Force SDL_WaitEventTimeout to 0 on bxt-rs side

khanghugo:
On parity with BunnymodXT to allow game to run in full performance in background, aka InActive.


YaLTeR:
Thanks

PR #52, opened 2022-10-28
Add bxt_viewmodel_remove

khanghugo:
I guess things can be added to this later down the avenue.

### DIFF
--- a/src/hooks/engine.rs
+++ b/src/hooks/engine.rs
@@ -536,6 +536,8 @@ pub static R_DrawSkyBox: Pointer<unsafe extern "C" fn()> = Pointer::empty_patter
     ]),
     my_R_DrawSkyBox as _,
 );
+pub static R_DrawViewModel: Pointer<unsafe extern "C" fn()> =
+    Pointer::empty_patterns(b"R_DrawViewModel\0", Patterns(&[]), my_R_DrawViewModel as _);
 pub static R_SetFrustum: Pointer<unsafe extern "C" fn()> = Pointer::empty_patterns(
     b"R_SetFrustum\0",
     // To find, search for "R_RenderView". This is R_RenderView(). The call between two if (global
@@ -843,6 +845,7 @@ static POINTERS: &[&dyn PointerTrait] = &[
     &R_Clear,
     &R_DrawSequentialPoly,
     &R_DrawSkyBox,
+    &R_DrawViewModel,
     &S_PaintChannels,
     &S_TransferStereo16,
     &SCR_DrawLoading,
@@ -1710,6 +1713,19 @@ pub mod exported {
             }
 
             R_DrawSkyBox.get(marker)();
+        })
+    }
+
+    #[export_name = "R_DrawViewModel"]
+    pub unsafe extern "C" fn my_R_DrawViewModel() {
+        abort_on_panic(move || {
+            let marker = MainThreadMarker::new();
+
+            if viewmodel::is_remove(marker) {
+                return;
+            }
+
+            R_DrawViewModel.get(marker)();
         })
     }
 
--- a/src/hooks/engine.rs
+++ b/src/hooks/engine.rs
@@ -540,7 +540,9 @@ pub static R_DrawViewModel: Pointer<unsafe extern "C" fn()> = Pointer::empty_pat
     b"R_DrawViewModel\0",
     Patterns(&[
         // 8684
-        pattern!(55 8B EC 83 EC 50 D9 05 ?? ?? ?? ?? D8 1D),
+        pattern!(55 8B EC 83 EC 50 D9 05 ?? ?? ?? ?? D8 1D ?? ?? ?? ?? 56 57 33 FF C7 45),
+        // 4554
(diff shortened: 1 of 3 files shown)


SmileyAG:
Can you also add 4554 pattern for `R_DrawViewModel`? We got found it before in **BunnymodXT**: https://github.com/YaLTeR/BunnymodXT/blob/1d9b3bdb6bf2fd3a1521404f9cf261a7f5ac0e33/BunnymodXT/patterns.hpp#L415-L419

src/hooks/engine.rs
```diff
@@ -536,6 +536,24 @@ pub static R_DrawSkyBox: Pointer<unsafe extern "C" fn()> = Pointer::empty_patter
     ]),
     my_R_DrawSkyBox as _,
 );
+pub static R_DrawViewModel: Pointer<unsafe extern "C" fn()> = Pointer::empty_patterns(
+    b"R_DrawViewModel\0",
+    Patterns(&[
```
YaLTeR, line 541: Every pointer with patterns must have a "// To find" comment, could you please add them according to what Smiley wrote? In the same style as other "// To find" commends.

src/modules/viewmodel.rs
```diff
@@ -0,0 +1,40 @@
+//! `bxt_viewmodel`
```
YaLTeR, line 1:
```suggestion
//! `bxt_viewmodel_remove`
```

src/modules/viewmodel.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use super::Module;
+use crate::hooks::engine;
+use crate::modules::cvars::CVar;
+use crate::utils::*;
+
+pub struct Viewmodel;
+impl Module for Viewmodel {
+    fn name(&self) -> &'static str {
+        "Viewmodel stuffs."
```
YaLTeR, line 11:
```suggestion
        "bxt_viewmodel_remove"
```

src/modules/viewmodel.rs
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    fn description(&self) -> &'static str {
+        "Do viewmodel stuffs."
```
YaLTeR, line 15:
```suggestion
        "Removing the viewmodel."
```
Please write proper strings next time, they show up on the wiki and in `bxt_help`. Other modules have examples.

src/modules/viewmodel.rs
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+static BXT_VIEWMODEL_REMOVE: CVar = CVar::new(
+    b"bxt_viewmodel_remove\0",
+    b"0\0",
+    "Set to `1` to disable rendering viewmodel.",
+);
+
+pub fn is_remove(marker: MainThreadMarker) -> bool {
```
YaLTeR, line 34:
```suggestion
pub fn is_removed(marker: MainThreadMarker) -> bool {
```
English

YaLTeR (changes requested):

src/modules/viewmodel.rs
khanghugo: Hmmm, I am just having the module named like that so maybe some more viewmodel stuff could be added in the future there. So you want this to be viewmodel_remove specific?

YaLTeR: Yes. The modules are split based on dependencies. In this case the module depends on `R_DrawViewModel` and `R_PreDrawViewModel`. If you added some more functionality here with extra dependencies, you would lose the ability to have one functionality enabled while the other functionality is disabled.

src/hooks/engine.rs
```diff
@@ -536,6 +536,33 @@ pub static R_DrawSkyBox: Pointer<unsafe extern "C" fn()> = Pointer::empty_patter
     ]),
     my_R_DrawSkyBox as _,
 );
+pub static R_DrawViewModel: Pointer<unsafe extern "C" fn()> = Pointer::empty_patterns(
+    // To find, search for "R_RenderView" function.
```
YaLTeR, line 540:
```suggestion
    // To find, search for "R_RenderView". This is R_RenderView.
```

src/hooks/engine.rs
```diff
@@ -536,6 +536,33 @@ pub static R_DrawSkyBox: Pointer<unsafe extern "C" fn()> = Pointer::empty_patter
...
+    Patterns(&[
+        // 8684
+        pattern!(55 8B EC 83 EC 50 D9 05 ?? ?? ?? ?? D8 1D ?? ?? ?? ?? 56 57 33 FF C7 45),
+        // 4554
+        pattern!(83 EC ?? D9 05 ?? ?? ?? ?? D8 1D ?? ?? ?? ?? 56 57 33 FF C7 44),
+    ]),
+    my_R_DrawViewModel as _,
+);
+pub static R_PreDrawViewModel: Pointer<unsafe extern "C" fn()> = Pointer::empty_patterns(
+    // To find, search for "R_RenderView" function.
```
YaLTeR, line 555:
```suggestion
    // To find, search for "R_RenderView". This is R_RenderView.
```

YaLTeR:
Thanks

PR #53, opened 2022-11-02
Add bxt_play_folder

khanghugo:
"wildcard" is not really wildcard as it won't attempt to match things like what it should, more meant for folder.

`enumerate()` pretty much does nothing but I think it is better to keep it there to log something for the tuple.

### DIFF
--- a/src/modules/demo_playback.rs
+++ b/src/modules/demo_playback.rs
@@ -22,7 +22,7 @@ impl Module for DemoPlayback {
     }
 
     fn commands(&self) -> &'static [&'static Command] {
-        static COMMANDS: &[&Command] = &[&BXT_PLAY_RUN];
+        static COMMANDS: &[&Command] = &[&BXT_PLAY_RUN, &BXT_PLAY_FOLDER];
         COMMANDS
     }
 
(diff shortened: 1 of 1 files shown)


src/modules/demo_playback.rs
```diff
@@ -134,16 +153,28 @@ fn find_demos(prefix: PathBuf) -> Result<Vec<(PathBuf, usize)>, String> {
         .filter_map(Result::ok)
         .map(|entry| entry.path())
         .filter(|path| path.extension().map(|ext| ext == "dem").unwrap_or(false))
-        .filter_map(|path| {
+        .enumerate()
+        .filter_map(|(index, path)| {
             path.file_stem()
                 .and_then(OsStr::to_str)
-                .filter(|name| name.starts_with(&name_prefix))
-                .and_then(|name| name[name_prefix.len()..].parse::<usize>().ok())
```
YaLTeR, line 141: Please don't add it into this function. The expression is already way too complicated, the last thing it needs is more special cases. Even just writing out a separate function with the same functionality would be better.

src/modules/demo_playback.rs
```diff
@@ -46,6 +46,16 @@ Plays back all `name_N.dem` demos in order.",
     ),
 );
 
+static BXT_PLAY_FOLDER: Command = Command::new(
+    b"bxt_play_folder\0",
+    handler!(
+        "bxt_play_folder <folder>
+        
+Plays back all demos in a folder in order.",
```
YaLTeR, line 54:
```suggestion
Plays back all demos in a folder in alphabetic order.",
```

YaLTeR (changes requested):

src/modules/demo_playback.rs
khanghugo: 👍

khanghugo: The `find_demos()` now has one for folder as requested 👍 I still have play_folder depending on play_run as it would be easier to maintain, I think.

YaLTeR:
Thanks! I'll deal with the filename limitation later because I'll need to check what the engine functions do again for that. For now I think it's decent enough, if you rename the folder to a single letter, you have 13 letters left for demo names.

khanghugo:
🐐

PR #59, opened 2022-12-03
demo_playback: Extend demo name length

khanghugo:


src/modules/demo_playback.rs
```diff
@@ -204,8 +204,18 @@ pub fn set_next_demo(marker: MainThreadMarker) {
...
+
+                // Assigning each row with sliced string
+                while len > 16 {
+                    cls_demos.demos[count].copy_from_slice(&demo[(count * 16)..((count + 1) * 16)]);
+                    len = len - 16;
+                    count = count + 1;
+                }
+                cls_demos.demos[count][..len]
+                    .copy_from_slice(&demo[(count * 16)..(count * 16 + len)]);
+                cls_demos.demos[count + 1][0] = 0;
```
khanghugo, line 218: Not sure how necessary this line 218 is. I take it out of my build and it still works. This line could cause segfault.

khanghugo: Thing is I don't think anyone would make a demo name 512 characters long so maybe this is fine until someone decides to just try it out.

src/modules/demo_playback.rs
```diff
@@ -204,8 +204,17 @@ pub fn set_next_demo(marker: MainThreadMarker) {
...
+                let mut len = demo.len();
+                let mut count = 0;
+
+                // Assigning each row with sliced string
+                while len > 16 {
+                    cls_demos.demos[count].copy_from_slice(&demo[(count * 16)..((count + 1) * 16)]);
+                    len = len - 16;
+                    count = count + 1;
+                }
+                cls_demos.demos[count][..len]
```
khanghugo, line 216: Maybe the second subscript can be culled also

YaLTeR:
Rust has a lot of iterators and encourages their use; you'll frequently get better optimized code when you use the iterators. Check how I implemented this.

khanghugo:
Yea, I was thinking of this but it won't look like that anyway. Thanks.

PR #61, opened 2023-01-30
add bxt_scoreboard_remove

khanghugo:
Some people asks for this and it is also a feature in HLAE

### DIFF
--- a/src/modules/scoreboard_remove.rs
+++ b/src/modules/scoreboard_remove.rs
@@ -1,0 +1,71 @@
+//! `bxt_scoreboard_remove`
+
+use super::Module;
+use crate::hooks::engine;
+use crate::modules::cvars::CVar;
+use crate::utils::*;
+
+pub struct ScoreboardRemove;
+impl Module for ScoreboardRemove {
+    fn name(&self) -> &'static str {
+        "bxt_scoreboard_remove"
+    }
+
+    fn description(&self) -> &'static str {
+        "Erasing `+showscores` in demo message"
+    }
+
+    fn cvars(&self) -> &'static [&'static CVar] {
+        static CVARS: &[&CVar] = &[&BXT_SCOREBOARD_REMOVE];
+        CVARS
+    }
+
+    fn is_enabled(&self, marker: MainThreadMarker) -> bool {
+        engine::Cbuf_AddText.is_set(marker)
+            || engine::Cbuf_AddFilteredText.is_set(marker)
+            || engine::Cbuf_AddTextToBuffer.is_set(marker)
+    }
+}
+
+static BXT_SCOREBOARD_REMOVE: CVar = CVar::new(
+    b"bxt_scoreboard_remove\0",
+    b"0\0",
+    "Set to `1` to hide scoreboard in demo.",
+);
+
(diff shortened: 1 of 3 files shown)


src/modules/scoreboard_remove.rs
```diff
@@ -0,0 +1,71 @@
...
+use crate::utils::*;
+
+pub struct ScoreboardRemove;
+impl Module for ScoreboardRemove {
+    fn name(&self) -> &'static str {
+        "bxt_scoreboard_remove"
+    }
+
+    fn description(&self) -> &'static str {
+        "Erasing `+showscores` in demo message"
```
YaLTeR, line 15:
```suggestion
        "Hiding the scoreboard by erasing the `+showscores` console command."
```

src/modules/scoreboard_remove.rs
```diff
@@ -0,0 +1,71 @@
...
+        engine::Cbuf_AddText.is_set(marker)
+            || engine::Cbuf_AddFilteredText.is_set(marker)
+            || engine::Cbuf_AddTextToBuffer.is_set(marker)
+    }
+}
+
+static BXT_SCOREBOARD_REMOVE: CVar = CVar::new(
+    b"bxt_scoreboard_remove\0",
+    b"0\0",
+    "Set to `1` to hide scoreboard in demo.",
```
YaLTeR, line 33:
```suggestion
    "Set to `1` to hide the scoreboard in demos.",
```

src/modules/scoreboard_remove.rs
```diff
@@ -0,0 +1,71 @@
...
+/// Returns `true` if scoreboard should currently be disabled.
+pub fn is_active(marker: MainThreadMarker) -> bool {
+    if !ScoreboardRemove.is_enabled(marker) {
+        return false;
+    }
+
+    BXT_SCOREBOARD_REMOVE.as_bool(marker)
+}
+
+pub unsafe fn on_cbuf_stuffs(marker: MainThreadMarker, text: *const i8) -> *const i8 {
```
YaLTeR, line 45: come on now, really? what is this function naming
```suggestion
pub unsafe fn strip_showscores(marker: MainThreadMarker, text: *const i8) -> *const i8 {
```

src/modules/scoreboard_remove.rs
```diff
@@ -0,0 +1,71 @@
...
+        for byte in a.bytes() {
+            if byte == *text_mut {
+                text_mut = text_mut.offset(1);
+            } else {
+                return text.cast();
+            }
+        }
+    }
+
+    text_mut.cast()
```
YaLTeR, line 70: Something like this should work better. Please test that this works, I didn't.
```suggestion
    let mut new_text: *const u8 = text.cast();
    for byte in b"+showscores" {
        if *new_text != *byte {
            // Character doesn't match, return original text.
            return text;
        }
        new_text = new_text.add(1);
    }

    new_text.cast()
```

YaLTeR (changes requested):

YaLTeR:
Please add back the comments as in my suggestion.

YaLTeR:
Thanks

PR #64, opened 2023-04-29
Add viewmodel sway

khanghugo:
~~Naming might be subjected to changes depending on uncool factors.~~

Name is changed. Only 6% less cool.

### DIFF
--- a/src/hooks/engine.rs
+++ b/src/hooks/engine.rs
@@ -121,6 +121,15 @@ pub static CL_PlayDemo_f: Pointer<unsafe extern "C" fn()> = Pointer::empty_patte
     ]),
     my_CL_PlayDemo_f as _,
 );
+// TODO: figure out how to use V_CalcRefdef from cl_funcs instead of this.
+pub static ClientDLL_CalcRefdef: Pointer<unsafe extern "C" fn(*mut ref_params_s)> =
+    Pointer::empty_patterns(
+        b"ClientDLL_CalcRefdef\0",
+        // You are in V_RenderView. The first call taking in one argument inside the first `do{}`
+        // block will be ClientDLL_CalcRefdef().
+        Patterns(&[]),
+        my_ClientDLL_CalcRefdef as _,
+    );
 pub static ClientDLL_Init: Pointer<unsafe extern "C" fn()> = Pointer::empty_patterns(
     b"ClientDLL_Init\0",
     // To find, search for "cl_dlls\\client.dll" (with a backslash).
@@ -144,6 +153,15 @@ pub static ClientDLL_DrawTransparentTriangles: Pointer<unsafe extern "C" fn()> =
         ]),
         null_mut(),
     );
+pub static cl: Pointer<*mut client_state_t> = Pointer::empty(b"cl\0");
+pub static cl_viewent: Pointer<*mut cl_entity_s> = Pointer::empty(
+    // Not a real symbol name.
+    b"cl_viewent\0",
+);
+pub static cl_viewent_viewmodel: Pointer<*mut cl_entity_s_viewmodel> = Pointer::empty(
+    // Not a real symbol name.
+    b"cl_viewent_viewmodel\0",
+);
 pub static cls: Pointer<*mut client_static_s> = Pointer::empty(b"cls\0");
 pub static cls_demos: Pointer<*mut client_static_s_demos> = Pointer::empty(
     // Not a real symbol name.
@@ -785,6 +803,17 @@ pub static V_FadeAlpha: Pointer<unsafe extern "C" fn() -> c_int> = Pointer::empt
     ]),
     my_V_FadeAlpha as _,
 );
+pub static V_RenderView: Pointer<unsafe extern "C" fn()> = Pointer::empty_patterns(
+    b"V_RenderView\0",
+    // To find, search for "R_RenderView: NULL worldmodel". This is an output error for
+    // R_RenderView(). Then, find a function calling R_RenderView() exactly 3 times. That
+    // function will be V_RenderView().
+    Patterns(&[
+        // 8684
+        pattern!(55 8B EC 81 EC F4 00 00 00 A1 ?? ?? ?? ?? 56 57),
+    ]),
+    null_mut(),
+);
 pub static VideoMode_IsWindowed: Pointer<unsafe extern "C" fn() -> c_int> = Pointer::empty_patterns(
     b"VideoMode_IsWindowed\0",
     // To find, first find GL_BeginRendering(). The first check is for the
@@ -845,7 +874,11 @@ static POINTERS: &[&dyn PointerTrait] = &[
     &CL_Move,
     &CL_PlayDemo_f,
     &ClientDLL_Init,
+    &ClientDLL_CalcRefdef,
     &ClientDLL_DrawTransparentTriangles,
+    &cl,
+    &cl_viewent,
+    &cl_viewent_viewmodel,
     &cls,
     &cls_demos,
     &Cmd_AddMallocCommand,
@@ -911,6 +944,7 @@ static POINTERS: &[&dyn PointerTrait] = &[
     &tri,
     &V_ApplyShake,
     &V_FadeAlpha,
+    &V_RenderView,
     &VideoMode_IsWindowed,
     &VideoMode_GetCurrentVideoMode,
     &window_rect,
@@ -949,7 +983,7 @@ pub struct ClientDllFunctions {
     pub CL_GetCameraOffsets: Option<NonNull<c_void>>,
     pub FindKey: Option<NonNull<c_void>>,
     pub CamThink: Option<NonNull<c_void>>,
-    pub CalcRefdef: Option<NonNull<c_void>>,
+    pub CalcRefdef: Option<unsafe extern "C" fn(*mut ref_params_s)>,
     pub AddEntity: Option<NonNull<c_void>>,
     pub CreateEntities: Option<NonNull<c_void>>,
     pub DrawNormalTriangles: Option<NonNull<c_void>>,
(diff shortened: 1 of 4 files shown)


src/hooks/engine.rs
```diff
@@ -780,6 +793,17 @@ pub static V_FadeAlpha: Pointer<unsafe extern "C" fn() -> c_int> = Pointer::empt
     ]),
     my_V_FadeAlpha as _,
 );
+pub static V_RenderView: Pointer<unsafe extern "C" fn()> = Pointer::empty_patterns(
+    b"V_RenderView\0",
+    // To find, search for "(0x4000)". This is the input for `glClear`. Cycle through results
```
YaLTeR, line 798: Like, text in disassembly? Sounds weird as a search target, maybe there's a more reasonable way?

src/hooks/engine.rs
```diff
@@ -962,6 +990,23 @@ pub struct dma_t {
     pub buffer: *mut c_uchar,
 }
 
+#[repr(C)]
+pub struct cl_entity_s {
+    pub index: c_int,
+    pub padding1: [c_char; 4 + 340 * 3],
+    pub current_position: c_int,
+    pub padding2: [c_char; 1792 + 8 + 52 + 4],
```
YaLTeR, line 998: This kind of long offset is likely different between engine versions, this is what the separate `cls_demos` type pointer is for.

src/hooks/engine.rs
```diff
@@ -962,6 +990,23 @@ pub struct dma_t {
...
+    pub origin: [c_float; 3],
+    pub angles: [c_float; 3],
+}
+
+#[repr(C)]
+pub struct client_state_t {
+    pub max_edicts: c_int,
+    pub padding1: [c_char; 1717500 - 4],
```
YaLTeR, line 1006: Same comment about offset being different, only in this case you did add separate `cl_viewent`?

src/hooks/engine.rs
```diff
@@ -1545,6 +1607,17 @@ pub unsafe fn find_pointers(marker: MainThreadMarker, base: *mut c_void, size: u
         _ => (),
     }
 
+    let ptr = &V_RenderView;
+    match ptr.pattern_index(marker) {
+        // 8684
+        Some(0) => {
+            // Find cl.viewent.index
```
YaLTeR, line 1614: Redundant comment?

src/modules/viewmodel_sway.rs
```diff
@@ -0,0 +1,127 @@
...
+        end
+    }
+}
+
+pub fn with_viewmodel_sway<T>(
+    marker: MainThreadMarker,
+    f: impl FnOnce() -> T,
+    rp: &ref_params_s,
+) -> T {
+    let rv = f();
```
YaLTeR, line 89: Call this in the hook and pass in rv for post-processing. No need to pass and call closure.

src/modules/viewmodel_sway.rs
```diff
@@ -0,0 +1,127 @@
...
+);
+
+static BXT_VIEWMODEL_SWAY_RATE: CVar = CVar::new(
+    b"bxt_viewmodel_sway_rate\0",
+    b"0.02\0",
+    "\
+The rate of sway.",
+);
+
+static LAST_VIEWANGLES: MainThreadCell<[f32; 3]> = MainThreadCell::new([0., 0., 0.]);
```
YaLTeR, line 54: Sounds like you need a `glam::Vec3`

src/modules/viewmodel_sway.rs
```diff
@@ -0,0 +1,127 @@
...
+    b"0\0",
+    "\
+Setting to `1` enables weapon sway. Default is `100` max and `0.02` rate.",
+);
+
+static BXT_VIEWMODEL_SWAY_MAX: CVar = CVar::new(
+    b"bxt_viewmodel_sway_max\0",
+    b"100\0",
+    "\
+The max of sway.",
```
YaLTeR, line 44: This is not how english works im pretty sure

src/modules/viewmodel_sway.rs
khanghugo: `ClientDLL_CalcRefdef` doesn't return anything but I just did that for consistency with others "with" functions... hehe

src/modules/viewmodel_sway.rs
khanghugo: I am too never fond of article.

src/hooks/engine.rs
khanghugo: To find, search for "R_RenderView: NULL worldmodel". This is an output error for `R_RenderView()`. Then, find a function calling `R_RenderView()` exactly 3 times. That function will be `V_RenderView()`.

Alternative it, I could mention the `glClear` input instead because there are three functions calling `R_RenderView()` but only `V_RenderView()` calling some functions with argument `0x4000`

src/modules/viewmodel_sway.rs
```diff
@@ -0,0 +1,119 @@
...
+use crate::utils::*;
+
+pub struct ViewmodelSway;
+impl Module for ViewmodelSway {
+    fn name(&self) -> &'static str {
+        "Viewmodel sway"
+    }
+
+    fn description(&self) -> &'static str {
+        "Viewmodel sway."
```
YaLTeR, line 16:
```suggestion
        "Adding CS:GO-like weapon sway."
```
or something

src/modules/viewmodel_sway.rs
```diff
@@ -0,0 +1,119 @@
...
+    "\
+Max of sway.",
+);
+
+static BXT_VIEWMODEL_SWAY_RATE: CVar = CVar::new(
+    b"bxt_viewmodel_sway_rate\0",
+    b"0.02\0",
+    "\
+Rate of sway.",
+);
```
YaLTeR, line 52: Are these actually needed? Like, why not just hardcode the numbers.

src/modules/viewmodel_sway.rs
```diff
@@ -0,0 +1,119 @@
...
+            start + (dabs * speed)
+        } else {
+            start - (dabs * speed)
+        }
+    } else {
+        end
+    }
+}
+
+pub fn with_viewmodel_sway(marker: MainThreadMarker, rp: &ref_params_s) {
```
YaLTeR, line 84:
```suggestion
pub fn add_viewmodel_sway(marker: MainThreadMarker, rp: &ref_params_s) {
```

src/modules/viewmodel_sway.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    let view = unsafe { &mut *engine::cl_viewent_viewmodel.get(marker) };
+    let mut last = LAST_VIEWANGLES.get(marker);
+
+    for i in 0..2 {
+        last[i] = normalize_angle(view.angles[i] - last[i]);
+    }
+
+    let m_x = last[1] * 0.045; // some numbers from the implementation
```
YaLTeR, line 100: What is this comment supposed to inform the reader about? Right now it does not inform me of anything, so might as well remove it.

src/modules/viewmodel_sway.rs
```diff
@@ -0,0 +1,119 @@
...
+    let m_y = last[0] * 0.045;
+    let frameadj = (1. / rp.frametime) * 0.01;
+    let max = BXT_VIEWMODEL_SWAY_MAX.as_f32(marker);
+    let vertical_reduction = (view.angles[0] / 180. * PI).cos().abs();
```
YaLTeR, line 104:
```suggestion
    let vertical_reduction = view.angles[0].to_radians().cos().abs();
```

src/modules/viewmodel_sway.rs
```diff
@@ -0,0 +1,119 @@
...
+pub fn with_viewmodel_sway(marker: MainThreadMarker, rp: &ref_params_s) {
+    if !BXT_VIEWMODEL_SWAY.as_bool(marker) {
+        return;
+    }
+
+    if rp.paused == 1 {
+        return;
+    }
+
+    let view = unsafe { &mut *engine::cl_viewent_viewmodel.get(marker) };
```
YaLTeR, line 93: Add a safety comment saying that no engine functions are called while the reference is active. You can search for existing SAFETY comments like this and copy the text.

src/hooks/engine.rs
```diff
@@ -2017,6 +2094,16 @@ pub mod exported {
         })
     }
 
+    #[export_name = "ClientDLL_CalcRefdef"]
+    pub unsafe extern "C" fn my_ClientDLL_CalcRefdef(rp: *mut ref_params_s) {
+        abort_on_panic(move || {
+            let marker = MainThreadMarker::new();
```
YaLTeR, line 2100:
```suggestion
            let marker = MainThreadMarker::new();

```

src/modules/viewmodel_sway.rs
```diff
@@ -0,0 +1,129 @@
...
+    view.angles[0] -= last[0];
+    view.angles[1] -= last[1];
+
+    for i in 0..3 {
+        view.origin[i] += BXT_VIEWMODEL_SWAY_RATE.as_f32(marker)
+            * vertical_reduction
+            * (-last[1].abs() * rp.forward[i] + last[0] * rp.up[i] + last[1] * rp.right[i]);
+    }
+
+    LAST_VIEWANGLES.set(marker, view.angles.into());
```
YaLTeR, line 128: So just last thing that caught my eye, shouldn't LAST_VIEWANGLES be reset to zero on disconnects/loads/whatever?

khanghugo: Sure but I don't think it would matter because it is set to something anew before being accessed to.

YaLTeR: I'm just thinking, if you spin wildly, then disconnect, load a completely unrelated demo, for the first frames you'll have the effects of the now unrelated spin?

khanghugo: Right, if it's worth it then I can add it some times later.

YaLTeR (approved):

YaLTeR:
Thanks

PR #67, opened 2023-06-09
Add campath

### LINKED ISSUE #62: HLAE camera path support
HLAE can import and export camera values from their own format .agr.
People can load the file in most 3D software rleady with HLAE own effort on supporting that.

Well, I just think if bxt-rs can load the file and record with sampling then bxt-rs can dilute a certain signature looking videomaking style.

All of the implementation should already been done in HLAE, my hope is that this would just be simple port.

khanghugo:
Closes #62 

Blender side must use HLAE Blender plugins: https://github.com/advancedfx/afx-blender-scripts.

The current state is very good despite some TODOs that I think are very niche. ~~I have not made anything with this to see if anything needs changing.~~

Lots of stuffs look verbose because I don't know Rust magic macros/built-in.

Both .Bvh and .Cam are supported is because .Bvh is more GoldSrc and .Cam is for Sourcn and not supported for GoldSrc by HLAE. I am just going extra to make sure anyone using it before could have it work here just fine. The difference is that .Bvh uses uniform frame time while .Cam uses timestamp and that works extremely well with Blender. It also has field of view for each entry so that enables some more effects.

The reason why we need export is to have a good reference of where the player is in Blender in order to aim better.

Default for offset is -0.25 because of loading time.


khanghugo:
The current implementation will override view even if current demo time is earlier than what is in the demo. I wasn't really thinking about this while making the other video. It should work like that because that is how it does when campath file ends. Will be an easy addition.

src/modules/campath/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+            && engine::scr_fov_value.is_set(marker)
+            && engine::V_RenderView.is_set(marker)
+    }
+}
+
+static BXT_CAMPATH_LOAD: CVar = CVar::new(
+    b"bxt_campath_load\0",
+    b"\0",
+    "\
+Loads campath .cam or .bvh file and playback during demo.
```
YaLTeR, line 63:
```suggestion
Loads campath .cam or .bvh file for playback during demo.
```

src/modules/campath/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+",
+);
+
+static BXT_CAMPATH_OFFSET: CVar = CVar::new(
+    b"bxt_campath_offset\0",
+    b"-0.25\0",
+    "\
+Offsets time values in second of each campath entry. Offset can be negative. Default is -0.25.
```
YaLTeR, line 71: > Default for offset is -0.25 because of loading time.

Why? Is this somehow not recorded during recording but played back during playback?

src/modules/campath/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+        );
+    }
+}
+
+static BXT_CAMPATH_EXPORT_STOP: Command = Command::new(
+    b"bxt_campath_export_stop\0",
+    handler!(
+        "bxt_campath_export_stop
+        
+Stops capturing player motion. Automatically stops when demo stops.",
```
YaLTeR, line 299:
```suggestion
Stops capturing player motion.",
```
If anything, this should go into export_start docs.

src/modules/campath/common.rs
```diff
@@ -0,0 +1,35 @@
...
+    (1. - t) * v0 + t * v1
+}
+
+pub fn angle_diff(a1: f64, a2: f64) -> f64 {
+    let a1 = a1.to_radians();
+    let a2 = a2.to_radians();
+
+    (a2.sin() * a1.cos() - a2.cos() * a1.sin())
+        .asin()
+        .to_degrees()
```
YaLTeR, line 34: What does this do? Why does it compute asin(sin(a2 - a1)) in a roundabout way?

src/modules/campath/camio.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        if time >= end.time {
+            return self.campaths.len();
+        }
+
+        // TODO: it can be more efficient
+        self.campaths
+            .iter()
+            .enumerate()
+            .find(|(_, e)| time < e.time)
```
YaLTeR, line 40: .position()

also you can combine it with condition above using .unwrap_or(self.campaths.len())

src/modules/campath/camio.rs
```diff
@@ -0,0 +1,143 @@
...
+            fov,
+            viewinfo: ViewInfo {
+                vieworg: [xpos, ypos, zpos],
+                viewangles: [yrot, zrot, xrot], // be mindful of the flip
+            },
+        },
+    )(i)
+}
+
+pub fn read_camio(i: &str) -> IResult<&str, CamIO> {
```
YaLTeR, line 131: Please add some tests for the parsing. Check hltas for example.

src/modules/campath/common.rs
```diff
@@ -0,0 +1,35 @@
+use nom::bytes::complete::tag;
+use nom::character::complete::space0;
+use nom::combinator::{map, opt, recognize};
+use nom::number::complete::float;
+use nom::sequence::preceded;
+use nom::IResult;
+
+#[derive(Clone, Copy)]
+pub struct ViewInfo {
+    pub vieworg: [f64; 3],
```
YaLTeR, line 10: If that helps, we should have Vec3 and similar types, they probably have lerp on them and stuff.

src/modules/campath/exporter.rs
```diff
@@ -0,0 +1,83 @@
...
+            .expect("Error: Cannot create .cam motion file.");
+        let mut file = BufWriter::new(file);
+
+        file.write_all(self.header().as_bytes())
+            .expect("Error: Cannot write motion into buffer.");
+        for idx in 0..self.data.campaths.len() {
+            file.write_all(self.entry_to_string(idx).as_bytes())
+                .expect("Error: Cannot write motion into buffer.");
+        }
+        file.flush()
```
YaLTeR, line 80: Why is this needed?

src/modules/campath/bvh.rs
```diff
@@ -0,0 +1,140 @@
...
+            cam_float, cam_float, cam_float, cam_float, cam_float, cam_float,
+        )), // parameters appear in the order of CHANNELS 6 specified in channel()
+        |(ypos, zpos, xpos, zrot, xrot, yrot)| ViewInfo {
+            vieworg: [-xpos, -ypos, zpos], // HLAE does this
+            viewangles: [-xrot, yrot, -zrot],
+        },
+    )(i)
+}
+
+pub fn read_bvh(i: &str) -> IResult<&str, Bvh> {
```
YaLTeR, line 125: Also please add parsing tests.

YaLTeR (changes requested):

src/modules/campath/mod.rs
khanghugo: The loading time is when you load the campath as soon as the demo/map starts. When you record a campath, it is not necessary that you are recording as soon as the map starts. So this offset only happens when loading demo. It is more than often that campath is played during demo playback, so this offset is for that.

src/modules/campath/common.rs
khanghugo: Smaller angle between two vectors. I did not know I could run this through WA (I am not major in intellectual field)

src/modules/campath/exporter.rs
khanghugo: Is that not how it works? I really have no idea how this bufwriter works so I guess it writes into a buffer (in memory) then I need to flush it to write to file. My intention is to write into memory before writing into file because in case of export 60fps or more (10-15fps is my usual to-go).

YaLTeR: I would expect it to flush on drop automatically. Can you check?

YaLTeR: Sorry, I should have checked the docs myself, too.

> It is critical to call [flush](https://doc.rust-lang.org/std/io/struct.BufWriter.html#method.flush) before BufWriter<W> is dropped. Though dropping will attempt to [flush](https://doc.rust-lang.org/std/io/struct.BufWriter.html#method.flush) the contents of the buffer, any errors that happen in the process of dropping will be ignored. Calling flush ensures that the buffer is empty and thus dropping will not even attempt file operations.

In that case it is indeed better to have an explicit flush() with expect(). Could you please add it back? Otherwise the PR looks good.

YaLTeR:
Thanks

PR #68, opened 2023-06-12
Add bxt_cap_time_*

khanghugo:
I was working with campath and sometimes I just want to capture a segment. Currently there is no way to scrub through demo quickly and this just seems to do the trick. Tested with normal and sampling capture.

Potentially, this could make resuming capture possible as well.

### DIFF
--- a/src/modules/capture/mod.rs
+++ b/src/modules/capture/mod.rs
@@ -6,7 +6,7 @@ use color_eyre::eyre::Context;
 
 use super::cvars::CVar;
 use super::{capture_video_per_demo, Module};
-use crate::hooks::engine::{self, con_print};
+use crate::hooks::engine::{self, con_print, prepend_command};
 use crate::modules::commands::Command;
 use crate::utils::*;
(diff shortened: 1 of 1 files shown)


khanghugo:
I am not sure how the `r_norefresh 1` still makes skipping demo very slow. It almost takes the same time as timedemo when I think it is supposed to be faster.

khanghugo:
This can be made as a different module I think. Or maybe it can stay vaporware.

src/modules/capture/mod.rs
```diff
@@ -281,6 +283,7 @@ pub fn cap_stop(marker: MainThreadMarker) {
     }
 
     capture_video_per_demo::stop(marker);
+    TIME.set(marker, 0.);
```
YaLTeR, line 286: Needs to be zeroed on cap start, not stop.

YaLTeR (commented):
Yea idk this does make sense, need to figure out how to best do it.

PR #69, opened 2023-06-30
Add missing Windows pattern

khanghugo:
I load the 25min TAS and it doesn't crash.

### DIFF
--- a/src/hooks/engine.rs
+++ b/src/hooks/engine.rs
@@ -43,7 +43,14 @@ pub static build_number: Pointer<unsafe extern "C" fn() -> c_int> = Pointer::emp
     null_mut(),
 );
 pub static CBaseUI__HideGameUI: Pointer<unsafe extern "fastcall" fn(*mut c_void)> =
-    Pointer::empty(b"_ZN7CBaseUI10HideGameUIEv\0");
+    Pointer::empty_patterns(
+        b"_ZN7CBaseUI10HideGameUIEv\0",
+        Patterns(&[
+            // 8684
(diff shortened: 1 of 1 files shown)


src/hooks/engine.rs
```diff
@@ -43,7 +43,14 @@ pub static build_number: Pointer<unsafe extern "C" fn() -> c_int> = Pointer::emp
     null_mut(),
 );
 pub static CBaseUI__HideGameUI: Pointer<unsafe extern "fastcall" fn(*mut c_void)> =
-    Pointer::empty(b"_ZN7CBaseUI10HideGameUIEv\0");
+    Pointer::empty_patterns(
+        b"_ZN7CBaseUI10HideGameUIEv\0",
+        Patterns(&[
```
YaLTeR, line 48: Any instruction how to find this?

khanghugo:
```
    // To find, search for "resource/trackerScheme.res". There will be two results and they are
    // inside CBaseUI::Start. The call with that string inside a conditional block will be Con_Printf().
    // Below the block will be an assignment to constant and strncpy() call. In between those two will be
    // (*gEngfuncs.pfnGetGameDirectory)() call that takes in zero arguments. 
    // (*gEngfuncs.pfnGetLevelName)() will be some bytes above the address of the other pointer and has
    // two references. The function with one argument and void return will be CBaseUI::HideGameUI().
 ```

Comment (khanghugo) [reply src/hooks/engine.rs:48]: Alternatively, 75th pointer of cl_enginefuncs.

Comment (khanghugo) [reply src/hooks/engine.rs:48]: Here is another alternative
```
    // 8th pointer in CBaseUI vtable.
    // To find, search for "chromehtml.dll". You are in CBaseUI::Initialize and that will be
    // the second pointer of CBaseUI vtable.
```

YaLTeR:
Thanks!

PR #71, opened 2023-07-10
editor: add constant yawspeed

khanghugo:


src/modules/tas_optimizer/optimizer.rs
```diff
@@ -510,9 +510,9 @@ fn mutate_frame_bulk<R: Rng>(rng: &mut R, frame_bulk: &mut FrameBulk) {
         dir: if strafe_type == StrafeType::MaxDeccel {
             StrafeDir::Best
         } else if rng.gen::<bool>() {
-            StrafeDir::Left
+            StrafeDir::Left(None)
```
YaLTeR, line 513: Could add mutating the turn rate just like with s03, I think it could be useful

src/modules/tas_studio/editor/mod.rs
```diff
@@ -2241,6 +2340,73 @@ impl Editor {
...
+        let from = String::from_utf8(buffer)
+            .expect("FrameBulk serialization should never produce invalid UTF-8");
+
+        let mut buffer = Vec::new();
+        hltas::write::gen_frame_bulk(&mut buffer, &new_bulk)
+            .expect("writing to an in-memory buffer should never fail");
+        let to = String::from_utf8(buffer)
+            .expect("FrameBulk serialization should never produce invalid UTF-8");
+
+        let op = Operation::Replace { line_idx, from, to };
```
YaLTeR, line 2406: If we switch to a separate strafing type as I mentioned in the hltas PR, then this function will only need to work on s5x frame bulks, and can use Operation::SetYawSpeed (which will be less taxing on the database size compared to Replace).

src/modules/tas_studio/mod.rs
```diff
@@ -1470,6 +1521,7 @@ fn add_frame_bulk_hud_lines(text: &mut Vec<u8>, bulk: &FrameBulk, last_frame: Op
         write!(text, "  X Pos: {:.1}\0", last_frame.state.player.pos.x).unwrap();
         write!(text, "  Y Pos: {:.1}\0", last_frame.state.player.pos.y).unwrap();
         write!(text, "  Z Pos: {:.1}\0", last_frame.state.player.pos.z).unwrap();
+        write!(text, "  Stamina: {:.1}\0", last_frame.state.player.stamina_time).unwrap();
```
YaLTeR, line 1524: Extract into separate PR, or honestly I'll just do this myself today

khanghugo: Go ahead with this one.

bxt-strafe/src/steps.rs
```diff
@@ -410,7 +410,21 @@ impl<S: Step> Step for Strafe<S> {
                     _ => 0.,
                 };
 
-                let vel_yaw = state.player.vel.y.atan2(state.player.vel.x);
+                let mut vel_yaw = state.player.vel.y.atan2(state.player.vel.x);
+
+                // Constant yawspeed.
+                // TODO: if yawspeed is too high, prediction will be wrong.
+                if let StrafeType::ConstYawspeed(yawspeed) = type_ {
+                    // Change vel_yaw in-place for convenience.
```
YaLTeR, line 418: So I'm looking at this code and thinking how to restructure it in a way that makes sense. And it got me thinking, what input is even expected with constant yawspeed strafing? Like, sure, the camera yaw should turn at a constant rate, but which input vector is expected? As an example, with vectorial strafing we can pick any other strafe type (max angle, max accel), then add a `change target_yaw to 180 over 1 s` and technically have constant camera yaw turn rate while the input will be anything else. This is certainly not what we expect though.

I guess the expected result here is always inputting 90 degrees to the left when turning left and 90 degrees to the right when turning right, equivalent to holding A and D button respectively.

Right now the code seems to set the velocity yaw as turned by the yaw delta, then apply a max accel theta to it. When testing with high aa especially, this will most of the time result in a 90 degree input vector indeed, however at lower speeds or aa, when theta is not 90 degrees, it will result in an input vector for which I don't see any reasonable meaning or explanation, so it will probably do something unexpected.

Therefore, I think for constant yawspeed, theta should be hardcoded to PI/2 (90 degrees).

As for a better way to structure this code, you could add an if branch across this part that computes and deals with vel_yaw and theta, and in the constant yawspeed part of the branch, instead of even using vel_yaw or theta, just `let camera_yaw = angle_mod_rad(input.yaw +- max_yaw_delta)` then set forward = 0 and side = +-constant whatever Vct sets it to.

YaLTeR: Might need a similar change in hlstrafe to avoid using theta for constyawspeed for similar reasoning.

src/modules/tas_studio/editor/mod.rs
```diff
@@ -1418,6 +1436,52 @@ impl Editor {
...
+            let op = Operation::SetYawspeed {
+                bulk_idx,
+                from: adjustment.original_value,
+                to: *yawspeed,
+            };
+            self.side_strafe_yawspeed_adjustment = None;
+            return self.store_operation(op);
+        }
+
+        // delta scalar might need being a lot higher but there's +ALT1 adjustment anyway.
```
YaLTeR, line 1471: I haven't tried this, but you can change the scale here to feel right. So that by default it's reasonable, and you might occasionally need shift or alt1. You shouldn't almost-always need to hold one of them.

src/modules/tas_studio/editor/mod.rs
```diff
@@ -1418,6 +1436,52 @@ impl Editor {
...
+        let speed = keyboard.adjustment_speed();
+        let delta = adjustment.delta(mouse.pos.as_vec2()) * 0.1 * speed;
+        let new_yawspeed = adjustment.original_value + delta;
```
YaLTeR, line 1474: Should this be maxed with 0? This brings another question, what happens if the yawspeed is negative? Does it even make sense?

src/modules/tas_studio/editor/operation.rs
```diff
@@ -519,6 +540,22 @@ impl Operation {
...
                 }
             }
+            Operation::SetYawspeed { bulk_idx, from, to } => {
+                let (bulk, first_frame_idx) = bulk_and_first_frame_idx_mut(hltas)
+                    .nth(bulk_idx)
+                    .expect("invalid bulk index");
+
+                let yawspeed = bulk
+                    .constant_yawspeed_mut()
+                    .expect("frame bulk should have max yawspeed even if empty");
```
YaLTeR, line 550: what is max yawspeed?

src/modules/tas_studio/editor/utils.rs
```diff
@@ -18,6 +18,12 @@ pub trait FrameBulkExt {
 
     /// Returns a mutable reference to the left-right count stored in the frame bulk, if any.
     fn left_right_count_mut(&mut self) -> Option<&mut NonZeroU32>;
+
+    /// Returns a reference to the yawspeed stored in the framebulk, if any.
+    fn constant_yawspeed(&self) -> Option<&f32>;
```
YaLTeR, line 23: I'd call them just `yawspeed` I think

src/modules/tas_studio/mod.rs
```diff
@@ -405,6 +406,28 @@ fn unset_yaw(marker: MainThreadMarker) {
     }
 }
 
+static BXT_TAS_STUDIO_SET_YAWSPEED: Command = Command::new(
+    b"bxt_tas_studio_set_yawspeed\0",
+    handler!(
+        "bxt_tas_studio_set_yawspeed
+        
+Sets the yawspeed of the selected frame bulk.",
```
YaLTeR, line 414:
```suggestion
Sets the yawspeed of the selected frame bulk for constant turn rate strafing.",
```

src/modules/tas_studio/editor/mod.rs
khanghugo: Yes, yawspeed can be negative and it strafes funny. I just don't want to deal with handling sign and extra mechanism so I allow yawspeed to be anything it wants. 0 is also a valid yawspeed.

PR #73, opened 2023-07-16
optimizer: add constant yawspeed

khanghugo:


src/modules/tas_optimizer/optimizer.rs
```diff
@@ -535,7 +535,31 @@ fn mutate_single_frame_bulk<R: Rng>(hltas: &mut HLTAS, rng: &mut R) -> usize {
...
             StrafeType::MaxAngle
         } else {
-            StrafeType::MaxAccel
+            // Constant yawspeed will not be selected unless specified in bulk.
+            if let StrafeType::ConstYawspeed(yawspeed) = *type_ {
+                StrafeType::ConstYawspeed(yawspeed)
+            } else {
+                StrafeType::MaxAccel
+            }
+        };
```
YaLTeR, line 544: This will let const yawspeed change to max deccel or max angle, but never to max accel, is that intentional?

khanghugo: Oh, definitely unintentional. I can make it more conditional.

src/modules/tas_optimizer/optimizer.rs
```diff
@@ -530,12 +530,36 @@ fn mutate_single_frame_bulk<R: Rng>(hltas: &mut HLTAS, rng: &mut R) -> usize {
...
+            // These numbers might be subjected to changes.
+            let scalar = if *yawspeed < 8. {
+                1.
+            } else if *yawspeed < 50. {
+                0.5
+            } else {
+                0.1
+            };
+
+            let range = (*yawspeed * scalar).max(5.);
```
YaLTeR, line 558: Is there any benefit to varying the range with yawspeed, rather than just using a constant range (i.e. +-5 always)? The optimizer can "climb" the small changes up, i.e. see how the s03 mutation works.

khanghugo: That would be nice. Let me try to do that and see. My idea is, as you see, higher yawspeed means it can only slowly increment while lower yawspeed, in realm of less than 2 digits, should change more drastically. If higher number has too wide range, probability will most likely produce bad results. If low number has too narrow range, changes might not be meaningful.

If it can "climb" the change then I can just do a constant +-0.5. The max yawspeed I have seen so far is around 600, I think.

YaLTeR:
Thanks

PR #75, opened 2023-07-19
optimizer: single frame/bulk refactor

khanghugo:


khanghugo:
This makes "mutate single frame bulk" to use "mutate frame bulk" so adding more stuffs doesn't need doing twice.

khanghugo:
I do see one difference is that it will not mutate if it is `AutoMovement::SetYaw` for mutate every single frame option. If that's worth it then I guess it can be changed.

YaLTeR:
So the reason single frame mutation is separate is because it is a special case which can be handled more efficiently. Specifically, when mutating a single frame, the optimizer never picks yaw strafing, because within a single frame you can only go left or right. So if you pick yaw strafing then you will have the big yaw angle search space wasting the RNG attempts, instead of the much simpler binary left or right choice. Therefore I'm not sure it's a good idea to join these two cases like that.

khanghugo:
Thanks. Understandable.

PR #83, opened 2023-08-02
Add bxt_emit_sound

khanghugo:
Edit: S_StartDynamicSound might be not what I want. It doesn't work per channel as I thought it would. The reason why I changed to S_StartDynamicSound is to avoid doing extra precache mechanism so any sounds can be played. I think I might need to revert two recent changes and then figure out how to do precache (or maybe not).

Edit 2: Never mind, it is good. I am just hallucinating.

Edit 3: After some more playing around, I think I have to make it similar to `speak` where it emits from where the player view is, rather than player origin. Big TODO in there and I thought it wouldn't come back.

Final edit: This should be good. I learn that the issue I've been having is just misusing the attenuation value. So now there are two options to emit sound with two different commands. Though for normal `bxt_emit_sound`, precaching might be a bit complicated for me for now to figure out how to do it properly. Meanwhile, `bxt_emit_sound_dynamic` can do that just fine. The difference between the two beside ability to play any sound is probably a jungle of sphagetti behind those twos. `bxt_emit_sound` will simply just work while `bxt_emit_sound_dynamic` might weirdly do something to other channels unless we pick correct channel.

### DIFF
--- a/src/modules/emit_sound.rs
+++ b/src/modules/emit_sound.rs
@@ -18,7 +18,7 @@ impl Module for EmitSound {
     }
 
     fn description(&self) -> &'static str {
-        "Exposing SV_StartSound to allow usage of other sound channels."
+        "Exposing some sound functions."
     }
 
     fn commands(&self) -> &'static [&'static Command] {
@@ -75,7 +75,7 @@ static BXT_EMIT_SOUND: Command = Command::new(
     handler!(
         "bxt_emit_sound <sound> <channel> [volume] [from] [to] [attenuation] [flag] [pitch]
 
-Plays sound file directly from SV_StartSound along with custom arguments.",
+Plays sound file directly from S_StartDynamicSound.",
         emit_sound as fn(_, _, _),
         emit_sound_full as fn(_, _)
     ),
--- a/src/modules/emit_sound.rs
+++ b/src/modules/emit_sound.rs
@@ -22,24 +22,29 @@ impl Module for EmitSound {
     }
 
     fn commands(&self) -> &'static [&'static Command] {
-        static COMMANDS: &[&Command] = &[&BXT_EMIT_SOUND];
+        static COMMANDS: &[&Command] = &[&BXT_EMIT_SOUND_STOP_SOUND_EXCEPT, &BXT_EMIT_SOUND];
         COMMANDS
     }
 
     fn is_enabled(&self, marker: MainThreadMarker) -> bool {
-        engine::SV_StartSound.is_set(marker) && engine::S_PrecacheSound.is_set(marker)
+        engine::S_StartDynamicSound.is_set(marker)
+            && engine::S_StopSound.is_set(marker)
+            && engine::S_PrecacheSound.is_set(marker)
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Default)]
+#[derive(Default)]
 struct SoundInfo {
     sound: String,
     channel: i32,
-    volume: i32,
-    entity_index: u32,
-    recipent: i32,
+    /// Volume is [0..1]
+    volume: f32,
+    from: i32,
+    to: i32,
+    /// "NORM" is 0.8
     attenuation: f32,
     flag: i32,
+    /// 100 is no shift.
     pitch: i32,
 }
 
@@ -54,9 +59,9 @@ impl FromStr for SoundInfo {
         let mut iter = s.split_ascii_whitespace();
         rv.sound = iter.next().unwrap_or_default().to_string();
         rv.channel = (f32::from_str(iter.next().unwrap_or_default())?).round() as i32;
-        rv.volume = (f32::from_str(iter.next().unwrap_or_default())?).round() as i32;
-        rv.entity_index = (f32::from_str(iter.next().unwrap_or_default())?).round() as u32;
-        rv.recipent = (f32::from_str(iter.next().unwrap_or_default())?).round() as i32;
+        rv.volume = f32::from_str(iter.next().unwrap_or_default())?;
+        rv.from = (f32::from_str(iter.next().unwrap_or_default())?).round() as i32;
+        rv.to = (f32::from_str(iter.next().unwrap_or_default())?).round() as i32;
         rv.attenuation = f32::from_str(iter.next().unwrap_or_default())?;
         rv.flag = (f32::from_str(iter.next().unwrap_or_default())?).round() as i32;
         rv.pitch = (f32::from_str(iter.next().unwrap_or_default())?).round() as i32;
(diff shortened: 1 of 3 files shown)


YaLTeR:
What's the usecase for these commands?

khanghugo:
To playback sound files very fast that it creates pitch. The way it is done now is to spam flashlight, weapon cycle, use button, speak commands. They only occupy 1-2 channels and the sounds come from the player. This PR exposes the functions and allows to play on different channel along with where the sound is emitted from. It is a less than legit way but it is harmless for any run.

YaLTeR:
Doesn't that make it even less legit than putting your own audio file and playing it with `speak`?

khanghugo:
This is purely performative. It can stay vaporwave here. I am fine with that.

YaLTeR:
Nah, it's fine to get it in. Was just wondering.

src/modules/emit_sound.rs
```diff
@@ -0,0 +1,219 @@
...
+use crate::utils::*;
+
+pub struct EmitSound;
+impl Module for EmitSound {
+    fn name(&self) -> &'static str {
+        "bxt_emit_sound"
+    }
+
+    fn description(&self) -> &'static str {
+        "Exposing some sound functions."
```
YaLTeR, line 21: Emitting and stopping sounds.

src/modules/emit_sound.rs
```diff
@@ -0,0 +1,219 @@
...
+            &BXT_EMIT_SOUND,
+        ];
+        COMMANDS
+    }
+
+    fn is_enabled(&self, marker: MainThreadMarker) -> bool {
+        engine::SV_StartSound.is_set(marker)
+            && engine::S_StartDynamicSound.is_set(marker)
+            && engine::S_StopSound.is_set(marker)
+            && engine::S_PrecacheSound.is_set(marker)
```
YaLTeR, line 37: listener_origin check is missing

src/modules/emit_sound.rs
```diff
@@ -0,0 +1,219 @@
...
+    }
+}
+
+#[derive(Default)]
```
YaLTeR, line 41: If this default is just for FromStr, then better remove it and in FromStr use the following pattern:
```
let volume = ...;
let from = ...;
...

Ok(SoundInfo {
    volume,
    from,
    ...
})
```

src/modules/emit_sound.rs
```diff
@@ -0,0 +1,219 @@
...
+
+fn emit_sound(marker: MainThreadMarker, sound: String, channel: i32) {
+    emit_sound_full(marker, SoundInfo::new(sound, channel));
+}
+
+fn emit_sound_full(marker: MainThreadMarker, info: SoundInfo) {
+    if let Some(player) = unsafe { engine::player_edict(marker) } {
+        let mut entity = unsafe { NonNull::new(player.as_ptr().add(info.from as usize)).unwrap() };
+
+        let entity = unsafe { entity.as_mut() };
```
YaLTeR, line 113: That is a bit of a roundabout way. I think you can do without the intermediate NonNull.

src/modules/emit_sound.rs
```diff
@@ -0,0 +1,219 @@
...
+impl FromStr for SoundInfo {
+    type Err = ParseFloatError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut rv = SoundInfo::default();
+
+        let mut iter = s.split_ascii_whitespace();
+        rv.sound = iter.next().unwrap_or_default().to_string();
+        rv.channel = (f32::from_str(iter.next().unwrap_or_default())?).round() as i32;
+        rv.volume = f32::from_str(iter.next().unwrap_or_default())?;
```
YaLTeR, line 84: Should this and perhaps other fields do clipping? Maybe at the point of invoking StartSound?

src/modules/emit_sound.rs
```diff
@@ -0,0 +1,219 @@
...
+                info.attenuation,
+                info.flag,
+                info.pitch,
+            )
+        };
+    }
+}
+
+static BXT_EMIT_SOUND_STOP_SOUND_EXCEPT: Command = Command::new(
+    b"bxt_emit_sound_stop_sound_except\0",
```
YaLTeR, line 183:
```suggestion
    b"bxt_stop_sound_except_channels\0",
```

src/modules/emit_sound.rs
```diff
@@ -0,0 +1,219 @@
...
+);
+
+// It goes up to 14-15 something. I've seen up to 8 used so far.
+static MAX_CHANNELS: i32 = 10;
+
+fn stop_sound(marker: MainThreadMarker, channels: String) {
+    let mut list = vec![];
+    // There should be only 8 usable channels from SV_StartSounds.
+    // But SND_PickDynamicChannel says something very weird otherwise.
+    let mut list0 = Vec::from_iter(0..MAX_CHANNELS + 1);
```
YaLTeR, line 199:
```suggestion
    let mut list0 = Vec::from_iter(0..=MAX_CHANNELS);
```

src/modules/emit_sound.rs
```diff
@@ -0,0 +1,219 @@
...
+            list.push(allow);
+        }
+    }
+
+    // Eh.
+    list.sort();
+
+    for channel in list.iter().rev() {
+        let channel = (*channel).clamp(0i32, MAX_CHANNELS);
+        list0.remove(channel as usize);
```
YaLTeR, line 213: You could instead collect individual channels into a HashSet, then do `for channel in 0..=MAX_CHANNELS {}` where call StopSound if the HashSet doesn't contain the index.

src/modules/emit_sound.rs
```diff
@@ -0,0 +1,219 @@
...
+    ),
+);
+
+fn emit_sound(marker: MainThreadMarker, sound: String, channel: i32) {
+    emit_sound_full(marker, SoundInfo::new(sound, channel));
+}
+
+fn emit_sound_full(marker: MainThreadMarker, info: SoundInfo) {
+    if let Some(player) = unsafe { engine::player_edict(marker) } {
+        let mut entity = unsafe { NonNull::new(player.as_ptr().add(info.from as usize)).unwrap() };
```
YaLTeR, line 111: Uh, no, reading from a pointer passed as a cvar argument with no checks is asking for weird crashes. This should only allow valid values.

src/modules/emit_sound.rs
```diff
@@ -0,0 +1,219 @@
...
+}
+
+fn emit_sound_dynamic_full(marker: MainThreadMarker, info: SoundInfo) {
+    if let Some(player) = unsafe { engine::player_edict(marker) } {
+        let origin = unsafe {
+            if info.to == 0 {
+                // It does this to have the sound follow player's vieworg rather than origin.
+                *engine::listener_origin.get(marker)
+            } else {
+                let to = NonNull::new(player.as_ptr().add(info.to as usize)).unwrap();
```
YaLTeR, line 158: Same here about unchecked input.

YaLTeR (changes requested):

src/modules/emit_sound.rs
khanghugo: I am not sure what this means.

YaLTeR: .clamp(0, 100) right here during parsing, or maybe at the point where SV_StartSound is called

src/modules/emit_sound.rs
```diff
@@ -0,0 +1,229 @@
...
+    ),
+);
+
+fn emit_sound(marker: MainThreadMarker, sound: String, channel: i32) {
+    emit_sound_full(marker, SoundInfo::new(sound, channel));
+}
+
+fn emit_sound_full(marker: MainThreadMarker, info: SoundInfo) {
+    if let Some(player) = unsafe { engine::player_edict(marker) } {
+        let entity = unsafe { player.as_ptr().add(info.from as usize).as_mut() };
```
YaLTeR, line 124: This should not use the player edict pointer as the base. It should use `server_t->edicts` and check `server_t->num_edicts` before accessing the pointer, as the edict count is dynamically allocated (and not statically bounded by 2047 as the `clamp`s above seem to imply). Something like this in BXT: https://github.com/YaLTeR/BunnymodXT/blob/71f1c9a75d6e92cb1f5e48c8d380541451165ea8/BunnymodXT/modules/HwDLL.hpp#L227-L230

src/modules/emit_sound.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+fn emit_sound_dynamic_full(marker: MainThreadMarker, info: SoundInfo) {
+    if let Some(player) = unsafe { engine::player_edict(marker) } {
+        let origin = unsafe {
+            if info.to == 0 {
+                // It does this to have the sound follow player's vieworg rather than origin.
+                *engine::listener_origin.get(marker)
+            } else {
+                let to = player.as_ptr().add(info.to as usize).as_ref();
```
YaLTeR, line 175: Same as above.

YaLTeR (changes requested):

src/modules/emit_sound.rs
```diff
@@ -120,34 +121,36 @@ fn emit_sound(marker: MainThreadMarker, sound: String, channel: i32) {
...
 fn emit_sound_full(marker: MainThreadMarker, info: SoundInfo) {
-    if let Some(player) = unsafe { engine::player_edict(marker) } {
-        let entity = unsafe { player.as_ptr().add(info.from as usize).as_mut() };
+    let entity = unsafe {
+        (*engine::sv_edicts.get(marker))
+            .add(info.from as usize)
+            .as_mut()
+    };
+
+    if entity.is_none() {
```
YaLTeR, line 130: This check does nothing. You're adding something to a pointer, it will never be NULL. You need to check num_edicts as I mentioned before.

src/modules/emit_sound.rs
```diff
@@ -0,0 +1,241 @@
...
+        COMMANDS
+    }
+
+    fn is_enabled(&self, marker: MainThreadMarker) -> bool {
+        engine::SV_StartSound.is_set(marker)
+            && engine::S_StartDynamicSound.is_set(marker)
+            && engine::S_StopSound.is_set(marker)
+            && engine::S_PrecacheSound.is_set(marker)
+            && engine::listener_origin.is_set(marker)
+            && engine::sv_edicts.is_set(marker)
```
YaLTeR, line 39:
```suggestion
            && engine::sv_edicts.is_set(marker)
            && engine::sv_num_edicts.is_set(marker)
```

src/modules/emit_sound.rs
```diff
@@ -0,0 +1,241 @@
...
+    emit_sound_dynamic_full(marker, SoundInfo::new(sound, channel))
+}
+
+fn emit_sound_dynamic_full(marker: MainThreadMarker, info: SoundInfo) {
+    let origin = unsafe {
+        if info.to == 0 {
+            // It does this to have the sound follow player's vieworg rather than origin.
+            *engine::listener_origin.get(marker)
+        } else {
+            if info.from >= *engine::sv_num_edicts.get(marker) {
```
YaLTeR, line 181:
```suggestion
            if info.to >= *engine::sv_num_edicts.get(marker) {
```

YaLTeR (changes requested):
I guess this is all that's left here?

khanghugo:
This one is just left here for more interesting videos like this https://www.youtube.com/watch?v=2q8zCNKMoPQ

But sure.

YaLTeR:
Thanks

PR #84, opened 2023-08-07
Add bxt_skyname

khanghugo:


khanghugo:
Oh, I did not know you made this before, sure, the name can be changed to that. Also, the reason why I have a command and a cvar for this is that having a cvar will properly persist change across levels.

SmileyAG:
> 

Oh, my purpose for that command was different, I used it for overriding skybox when starting the demo playback (since `sv_skyname` doesn't do it)

khanghugo:
It looks like your implementation should work for normal gameplay as well as demo because it looks exactly like mine. Also for demo playback but alsofor  mappers I guess because they requested it.

What do you think of naming it `bxt_skybox...` because there's also `bxt_skybox_remove` in bxt-rs? And what do you think of the ability to have skyname set to that cvar every map? I mean I can make the command to optionally take another argument but that would modify the cvar and I have not seen any commands freely setting other cvar so that makes doing it very iffy.

src/hooks/engine.rs
```diff
@@ -1939,6 +1997,19 @@ pub mod exported {
         })
     }
 
+    #[export_name = "R_LoadSkys"]
+    pub unsafe extern "C" fn my_R_LoadSkys() {
+        abort_on_panic(move || {
+            let marker = MainThreadMarker::new();
+
+            skybox_change::change_name(marker);
```
YaLTeR, line 2005: A with_changed_name() function would be good here, see some other with_ in some other hooked function

src/modules/skybox_change.rs
```diff
@@ -0,0 +1,106 @@
...
+    }
+}
+
+static BXT_SKYNAME: CVar = CVar::new(
+    b"bxt_skyname\0",
+    b"\0",
+    "\
+Changes skybox name. This does not take effect instantaneously unless force command is also invoked.
+However, if specified before map load, it will take effect.
+Eg: `bxt_skyname city`",
```
YaLTeR, line 45:
```suggestion
E.g.: `bxt_skyname city`",
```

src/modules/skybox_change.rs
```diff
@@ -0,0 +1,106 @@
...
+            && engine::R_LoadSkys.is_set(marker)
+            && engine::gLoadSky.is_set(marker)
+    }
+}
+
+static BXT_SKYNAME: CVar = CVar::new(
+    b"bxt_skyname\0",
+    b"\0",
+    "\
+Changes skybox name. This does not take effect instantaneously unless force command is also invoked.
```
YaLTeR, line 43:
```suggestion
Sets skybox name.

This does not take effect instantaneously unless force command is also invoked.
```

src/modules/skybox_change.rs
```diff
@@ -0,0 +1,106 @@
...
+pub fn change_name(marker: MainThreadMarker) {
+    if !SkyboxChange.is_enabled(marker) {
+        return;
+    }
+
+    let mv = unsafe { &mut *engine::movevars.get(marker) };
+    let bytes = BXT_SKYNAME.to_string(marker).into_bytes();
+    let original_skyname = *ORIGINAL_SKYNAME.borrow_mut(marker);
+
+    if original_skyname[0] == 0 {
```
YaLTeR, line 69: Should this be an assert_eq!(original_skyname[0], 0)?

src/modules/skybox_change.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    mv.skyName = *ORIGINAL_SKYNAME.borrow(marker);
+}
+
+fn force(marker: MainThreadMarker) {
+    change_name(marker);
+
+    unsafe {
+        // One single boolean check.
+        *engine::gLoadSky.get(marker) = 1;
```
YaLTeR, line 97: Should this be reset to 0 after?

YaLTeR (commented):
> What do you think of naming it bxt_skybox... because there's also bxt_skybox_remove in bxt-rs

Makes sense, bxt_skybox_name and bxt_skybox_reload (instead of force) maybe

src/hooks/engine.rs
khanghugo: I'm curious about this design. What happens if we have another functionality involving something like that? Wouldn't we have to undo this? I won't try to think too much about it

src/modules/skybox_change.rs
khanghugo: This boolean check is sky load sky draw init something. That function is 1 line changing `gLoadSky` to 1. Instead of hooking the function, I hook the variable and change it directly. R_LoadSkys` sets the variable to 1 after the it finishes itself.

src/hooks/engine.rs
khanghugo: Oh, never mind, Im stupid, it makes sense.

src/modules/skybox_change.rs
khanghugo: Hmmm, I get what you mean now. Let me try to make it not crash in anyway.

src/modules/skybox_change.rs
```diff
@@ -0,0 +1,114 @@
...
+    b"bxt_skybox_name\0",
+    b"\0",
+    "\
+Sets skybox name.
+
+This does not take effect instantaneously unless reload command is also invoked.
+E.g.: `bxt_skybox_name city`",
+);
+
+static BXT_SKYNAME_RELOAD: Command = Command::new(
```
YaLTeR, line 49:
```suggestion
static BXT_SKYBOX_RELOAD: Command = Command::new(
```

src/modules/skybox_change.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+static BXT_SKYBOX_NAME: CVar = CVar::new(
+    b"bxt_skybox_name\0",
+    b"\0",
+    "\
+Sets skybox name.
+
+This does not take effect instantaneously unless reload command is also invoked.
+E.g.: `bxt_skybox_name city`",
```
YaLTeR, line 46:
```suggestion

Example: `bxt_skybox_name city`",
```

src/modules/skybox_change.rs
```diff
@@ -0,0 +1,114 @@
...
+    unsafe {
+        // One single boolean check.
+        *engine::gLoadSky.get(marker) = 1;
+        engine::R_LoadSkys.get(marker)();
+    }
+
+    restore(marker);
+}
+
+fn reset_skyname(marker: MainThreadMarker) {
```
YaLTeR, line 112: It's used only once in restore() right? Might as well just inline it there.

src/modules/skybox_change.rs
```diff
@@ -0,0 +1,114 @@
...
+fn restore(marker: MainThreadMarker) {
+    // Demo playback needs this. Though very unnecessary, it is nice to have.
+    let mv = unsafe { &mut *engine::movevars.get(marker) };
+
+    mv.skyName = *ORIGINAL_SKYNAME.borrow(marker);
+    reset_skyname(marker)
+}
+
+fn force(marker: MainThreadMarker) {
+    change_name(marker);
```
YaLTeR, line 101: Let's also use the with_ function here

PR #85, opened 2023-09-02
Port bxt_fix_widescreen_fov

khanghugo:
Port of https://github.com/YaLTeR/BunnymodXT/pull/448 with fix from https://github.com/YaLTeR/BunnymodXT/issues/451

I was thinking of doing the fix differently by restoring `scr_fov_value` after some functions are done using it (biggestly `R_RenderView`).

That might be or might be not necessary. Because this is a "fix", so it should not really write some info into the demo, I thought to myself. But I guess this would be just fine.

### DIFF
--- a/src/modules/fix_widescreen.rs
+++ b/src/modules/fix_widescreen.rs
@@ -16,7 +16,11 @@ impl Module for FixWidescreen {
     }
 
     fn is_enabled(&self, marker: MainThreadMarker) -> bool {
-        engine::R_SetFrustum.is_set(marker) && engine::scr_fov_value.is_set(marker)
+        engine::R_SetFrustum.is_set(marker)
+            && engine::scr_fov_value.is_set(marker)
+            && engine::VideoMode_IsWindowed.is_set(marker)
+            && engine::VideoMode_GetCurrentVideoMode.is_set(marker)
+            && engine::window_rect.is_set(marker)
     }
 
     fn cvars(&self) -> &'static [&'static CVar] {
@@ -39,7 +43,7 @@ Fixes reduction in vertical field-of-view of widescreen. Compatible with `bxt_fo
 // YaLTeR: It's set in the client DLL in HUD_something which runs every frame but not during loads.
 static PREV_FOV: MainThreadRefCell<f32> = MainThreadRefCell::new(0.);
 
-pub unsafe fn fix_widescreen_fov(marker: MainThreadMarker) {
+pub fn fix_widescreen_fov(marker: MainThreadMarker) {
     if !FixWidescreen.is_enabled(marker) {
         return;
     }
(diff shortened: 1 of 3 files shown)


src/modules/fix_widescreen.rs
```diff
@@ -0,0 +1,70 @@
...
+        static CVARS: &[&CVar] = &[&BXT_FIX_WIDESCREEN_FOV];
+        CVARS
+    }
+}
+
+static BXT_FIX_WIDESCREEN_FOV: CVar = CVar::new(
+    b"bxt_fix_widescreen_fov\0",
+    b"0\0",
+    "\
+Fixes reduction in vertical field-of-view of widescreen. Compatible with `bxt_force_fov`",
```
YaLTeR, line 32:
```suggestion
Fixes reduction in vertical field-of-view of widescreen. Compatible with `bxt_force_fov`.",
```

src/modules/fix_widescreen.rs
```diff
@@ -0,0 +1,70 @@
...
+use crate::utils::*;
+
+pub struct FixWidescreen;
+impl Module for FixWidescreen {
+    fn name(&self) -> &'static str {
+        "bxt_fix_widescreen_fov"
+    }
+
+    fn description(&self) -> &'static str {
+        "Fixes widescreen vertical field-of-view."
```
YaLTeR, line 15:
```suggestion
        "Correcting widescreen vertical field-of-view."
```

src/modules/fix_widescreen.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+static BXT_FIX_WIDESCREEN_FOV: CVar = CVar::new(
+    b"bxt_fix_widescreen_fov\0",
+    b"0\0",
+    "\
+Fixes reduction in vertical field-of-view of widescreen. Compatible with `bxt_force_fov`",
+);
+
+// My guess is that every render cycle, `scr_fov_value` is reset to some values.
```
YaLTeR, line 35: It's set in the client DLL in HUD_something which runs every frame but not during loads.

src/modules/fix_widescreen.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    if !BXT_FIX_WIDESCREEN_FOV.as_bool(marker) {
+        return;
+    }
+
+    let fov = unsafe { *engine::scr_fov_value.get(marker) };
+    let prev_fov = *PREV_FOV.borrow(marker);
+
+    if fov != prev_fov {
+        let (width, height) = unsafe { engine::get_resolution(marker) };
```
YaLTeR, line 53: Is this get_resolution() covered in is_enabled()?

YaLTeR (changes requested):

src/modules/fix_widescreen.rs
khanghugo: yea I think this makes sense because it does not really fix anything as it also changes horizontal FOV. For a while I thought that is how it works because it just looks wide.

src/modules/fix_widescreen.rs
khanghugo: I see, I guess I will just have the keyword unsafe for the function

YaLTeR: I mean, just add the checks for whatever get_resolution uses into is_enabled

YaLTeR:
Thanks

PR #86, opened 2023-09-20
studio: fix CS 1.6 unfocused playback

khanghugo:
More explanation.

In Bunnnymodxt, HwDLL.cpp, this snippet is executed.

image

In bxt-rs playback hook `bxt_tas_load_script_from_string`, that part is not included.

The important part here is setting active mouse.


khanghugo:
I am not sure I should include it in the simulator game part but the simulation is so far seems fine for me. Last time I looked at it, the simulator got the accurate pitch just fine. But then I remember at some points, it got wrong. Could be Mandela's effect. But this should be just fine.

YaLTeR:
Thanks

PR #87, opened 2023-09-24
studio: hovered frame last frame of selected bulk during adjustment

khanghugo:


YaLTeR (commented):
Maybe it's better to put all this logic into editor.hovered_frame_idx()? Is that gonna break anything?

khanghugo:
Sure. Ill do that

khanghugo:
In camera editor, there is no selected bulk so hovered frame will be whatever closer to the mouse. I don't think it is worth making it work with camera editor because it is not very important.

khanghugo:
I think I discovered a very weird unrelated bug from this.

I have tested this and I made a whole TAS with this build. However, I recently just started a TAS and it crashes when I try to adjust the final segment. The crash is index out of bound. It is very odd considering that the reported crash index is there in the frame but it says it is out of bound. So as long as I adjust with right click, it will crash every time. Once I extend the final segment, it works normally.

So my guess is that for some reasons the branch frames are not updated despite what I see on screen until I make another change that would change the number of frames.

PR #91, opened 2024-01-05
Add missing capture skip non gameplay condition

khanghugo:
Merge issue with other branches because this line is in the PR commit.


YaLTeR:
Thanks

PR #94, opened 2024-01-29
Fix capture skip non gameplay frame

### LINKED ISSUE #93: [BUG] "bxt_cap_skip_non_gameplay_frames 1" crashes when demo(s) enqueued for playback
Bug found with latest bxt-rs-Windows-release artifact built off master

Reproduction steps:
1. Start game with bxt-rs injected (game itself doesn't appear to matter - originally found with GMC, but reproduced the issue in HL as well).
2. Execute command `bxt_cap_skip_non_gameplay_frames 1`.
3. Enqueue demo(s) for playback, either via `playdemo` or via `bxt_play_run`/`bxt_play_folder`.

Game crashes after freezing briefly on an ostensible OOB frame from the demo.

Same crash can be reproduced via `bxt_play_run`/`bxt_play_folder` then `bxt_cap_skip_non_gameplay_frames 1` during demo playback, but _not_ via `playdemo` on a single demo then `bxt_cap_skip_non_gameplay_frames 1` while it plays back, indicating the issue is with demos being queued to play rather than demos playing outright.

Note that `bxt_cap_start` does not need to be executed in order to reproduce the crash.

When using `bxt_play_run`/`bxt_play_folder`, bxt-rs.log shows DEBUG lines for demos being queued for playback, but nothing after that.

khanghugo:
Closes #93 

might be a terrible work around. Let me seee if I can do something different.

### DIFF
--- a/src/modules/capture/mod.rs
+++ b/src/modules/capture/mod.rs
@@ -200,13 +200,13 @@ pub enum SoundCaptureMode {
 }
 
 #[allow(clippy::large_enum_variant)]
-enum State {
+pub enum State {
     Idle,
     Starting(String),
     Recording(Recorder),
 }
 
-static STATE: MainThreadRefCell<State> = MainThreadRefCell::new(State::Idle);
+pub static STATE: MainThreadRefCell<State> = MainThreadRefCell::new(State::Idle);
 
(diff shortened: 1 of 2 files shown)


khanghugo:
probably unrelated

PR #98, opened 2024-03-29
Add pitch smoothing

khanghugo:


khanghugo:
Waiting for hltas update.

I think it is okay if it just works for global smoothing but not individual hovered segment.

src/modules/tas_studio/editor/mod.rs
```diff
@@ -3879,43 +3895,56 @@ fn unwrap_angles(xs: impl Iterator<Item = f32>) -> impl Iterator<Item = f32> {
     })
 }
 
-fn smoothed_yaws(
```
YaLTeR, line 3882: Instead of doing the same operations twice on two values (error-prone), pass the angles to work with explicitly. For example, move `let yaws = frames.iter().map(|f| f.state.prev_frame_input.yaw);` outside the function and accept it as an argument. Then the function can be called separately for pitch and yaw as needed.

src/modules/tas_studio/editor/mod.rs
```diff
@@ -2841,14 +2841,23 @@ impl Editor {
...
-        for yaw in &smoothed[1..] {
+        for yaw in &smoothed.1[1..] {
             let yaw = yaw.to_degrees();
             write!(&mut line, " {yaw}").unwrap();
         }
 
         let op = Operation::Insert { line_idx: 0, line };
         self.apply_operation(op)?;
 
+        let mut line = "pitch_override".to_string();
```
YaLTeR, line 2852: If the pitch was constant then it probably doesn't need to be smoothed? Also do you generally want to smooth both at once?

khanghugo: I am not sure how I would change existing functions to stop working for stream of constant pitch. It just seems a lot easier to mirror it and just works. I see what you mean about second point where someone might just want to smooth the yaw. However, they can delete it easily. Also, adding another option to smooth pitch might be inconvenient as if soneone wants smoothing, just smooth it.

src/modules/tas_studio/editor/mod.rs
```diff
@@ -3974,14 +4019,14 @@ fn smoothed_yaws(
 
         // Walk forward half an interval.
         middle_frame_half
-            .chain(zip(
+            .chain(izip!(
```
YaLTeR, line 4022:
```suggestion
            .chain(zip(
```
This can be rolled back. And import removed.

YaLTeR (approved):

khanghugo:
If anything can merge https://github.com/YaLTeR/bxt-rs/pull/101 before merging this.

khanghugo:
Rebased. I tested with one long TAS run I have and there's no problems.

YaLTeR:
Thanks for testing.

PR #99, opened 2024-03-30
Add bxt_tas_studio_replay_views

### LINKED ISSUE #97: TAS Editor Camera Replay
Instead of replaying the whole TAS, there could be an alternative mode where it replays the views instead. Much faster and okayish enough for things.

khanghugo:
Closes #97 

Intended to work with yaw (and pitch) override and also without.

Seems pretty okay.

I don't want to figure out how to keep the same position after replay but if it is needed then I shall.

If no second game and no tas playback first time then viewangles replay won't work (not a problem and no one would encounter it but can be fixed very easy or not depending on the simulated run would override the values or not).

And
> bxt_tas_norefresh_until_last_frames = 0 means 1 frame being played. Heh.

If that needs changing then sure. I don't want to complicate all of the logic too much.

### DIFF
--- a/src/hooks/bxt.rs
+++ b/src/hooks/bxt.rs
@@ -141,6 +141,7 @@ pub struct OnTasPlaybackFrameData {
     pub strafe_cycle_frame_count: u32,
     pub prev_predicted_trace_fractions: [f32; 4],
     pub prev_predicted_trace_normal_zs: [f32; 4],
+    pub rendered_viewangles: [f32; 3],
 }
 
 unsafe extern "C" fn on_tas_playback_frame(data: OnTasPlaybackFrameData) -> c_int {
--- a/src/hooks/engine.rs
+++ b/src/hooks/engine.rs
@@ -723,6 +723,14 @@ pub static SCR_DrawLoading: Pointer<unsafe extern "C" fn()> = Pointer::empty_pat
     ]),
     my_SCR_DrawLoading as _,
 );
+pub static SCR_DrawPause: Pointer<unsafe extern "C" fn()> = Pointer::empty_patterns(
+    b"SCR_DrawPause\0",
+    // To find, search for "cz_worldmap". You are in SCR_DrawPause().
+    Patterns(&[
+        pattern!(D9 05 ?? ?? ?? ?? D8 1D ?? ?? ?? ?? DF E0 F6 C4 44 7B ?? A1 ?? ?? ?? ?? 85 C0 74 ?? E8 ?? ?? ?? ?? 85),
+    ]),
+    my_SCR_DrawPause as _,
+);
 pub static scr_fov_value: Pointer<*mut c_float> = Pointer::empty(b"scr_fov_value\0");
 pub static shm: Pointer<*mut *mut dma_t> = Pointer::empty(b"shm\0");
 pub static sv: Pointer<*mut c_void> = Pointer::empty(b"sv\0");
@@ -987,6 +995,7 @@ static POINTERS: &[&dyn PointerTrait] = &[
     &S_PaintChannels,
     &S_TransferStereo16,
     &SCR_DrawLoading,
+    &SCR_DrawPause,
     &scr_fov_value,
     &shm,
     &sv,
@@ -2191,6 +2200,19 @@ pub mod exported {
         })
     }
 
+    #[export_name = "SCR_DrawPause"]
+    pub unsafe extern "C" fn my_SCR_DrawPause() {
+        abort_on_panic(move || {
+            let marker = MainThreadMarker::new();
+
+            if !tas_studio::should_draw_pause(marker) {
+                return;
+            }
+
+            SCR_DrawPause.get(marker)();
+        })
+    }
+
     #[export_name = "SV_Frame"]
     pub unsafe extern "C" fn my_SV_Frame() {
         abort_on_panic(move || {
@@ -2376,6 +2398,7 @@ pub mod exported {
             let marker = MainThreadMarker::new();
 
             campath::override_view(marker);
+            tas_studio::tas_playback_rendered_views(marker);
 
             R_RenderView.get(marker)();
         })
--- a/src/modules/tas_studio/mod.rs
+++ b/src/modules/tas_studio/mod.rs
@@ -71,6 +71,7 @@ impl Module for TasStudio {
             &BXT_TAS_STUDIO_NEW,
             &BXT_TAS_STUDIO_LOAD,
             &BXT_TAS_STUDIO_CONVERT_HLTAS,
+            &BXT_TAS_STUDIO_REPLAY_VIEWS,
             &BXT_TAS_STUDIO_REPLAY,
             &BXT_TAS_STUDIO_SET_STOP_FRAME,
             &BXT_TAS_STUDIO_SET_YAWSPEED,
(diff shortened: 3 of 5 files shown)


khanghugo:
Maybe this could be renamed to `bxt_tas_studio_replay_camera`. Whatever.

src/modules/tas_studio/mod.rs
```diff
@@ -1560,7 +1638,82 @@ pub unsafe fn maybe_receive_messages_from_remote_server(marker: MainThreadMarker
...
+
+                r_refdef_vieworg[0] = editor.branch().frames[current_frame].state.player.pos[0];
+                r_refdef_vieworg[1] = editor.branch().frames[current_frame].state.player.pos[1];
+                r_refdef_vieworg[2] = editor.branch().frames[current_frame].state.player.pos[2];
+
+                // We don't keep track of vieworg so just deduce it from here instead.
+                // It is not easy to keep track of vieworg because it isn't there to track.
+                // vieworg is calculated.
+                // It is also funny how vieworg is calculated here,
+                // but viewangles is passed from BXT.
```
YaLTeR, line 1691: I was gonna ask why this is necessary but I guess bxt-strafe just doesn't handle angle overrides yet?

Ideally it would handle overrides (then they will also work without simulation), but I guess that's effort?

khanghugo: Yea I'd say bxt-strafe is not aware of the viewangles / rendered angle data so you might call this precursor to that. This is just half of the work if the simulation wants the overrides. However, this one is specifically just the rendered yaw, not the actual player yaw.

src/hooks/engine.rs
```diff
@@ -2376,6 +2398,7 @@ pub mod exported {
             let marker = MainThreadMarker::new();
 
             campath::override_view(marker);
+            tas_studio::tas_playback_rendered_views(marker);
```
YaLTeR, line 2401: Should this be before or after campath?

src/modules/tas_studio/mod.rs
```diff
@@ -433,6 +436,71 @@ fn replay(marker: MainThreadMarker) {
...
+        } => {
+            editor.cancel_ongoing_adjustments();
+
+            sdl::set_relative_mouse_mode(marker, true);
+            client::activate_mouse(marker, true);
+
+            // bxt_tas_norefresh_until_last_frames = 0 means 1 frame being played. Heh.
+            let start_frame = norefresh_until_stop_frame_frame_idx(marker, &editor);
+
+            engine::prepend_command(marker, "bxt_hud 0; hud_draw 0\n");
```
YaLTeR, line 467: Do we want this?

src/modules/tas_studio/mod.rs
```diff
@@ -433,6 +436,71 @@ fn replay(marker: MainThreadMarker) {
...
+
+            State::Editing {
+                editor,
+                last_generation,
+                last_branch_idx,
+                simulate_at,
+                bridge,
+            }
+        }
+        other => other,
```
YaLTeR, line 500: Print a message that you need to be in the editor for it to work

src/modules/tas_studio/mod.rs
```diff
@@ -1560,7 +1638,82 @@ pub unsafe fn maybe_receive_messages_from_remote_server(marker: MainThreadMarker
             editor.recompute_extra_camera_frame_data_if_needed();
         }
         State::PreparingToPlayToEditor(_, _, _) => unreachable!(),
+        _ => (),
```
YaLTeR, line 1641: Shouldn't the playing views state be added to the PlayingToEditor and Editing arm above?

src/modules/tas_studio/mod.rs
```diff
@@ -1560,7 +1638,82 @@ pub unsafe fn maybe_receive_messages_from_remote_server(marker: MainThreadMarker
...
+                    simulate_at,
+                    bridge,
+                }
+            } else {
+                let r_refdef_vieworg = unsafe { &mut *engine::r_refdef_vieworg.get(marker) };
+                let r_refdef_viewangles = unsafe { &mut *engine::r_refdef_viewangles.get(marker) };
+
+                r_refdef_vieworg[0] = editor.branch().frames[current_frame].state.player.pos[0];
+                r_refdef_vieworg[1] = editor.branch().frames[current_frame].state.player.pos[1];
+                r_refdef_vieworg[2] = editor.branch().frames[current_frame].state.player.pos[2];
```
YaLTeR, line 1685: Extract into variable `&editor.branch().frames[current_frame]` or even `&editor.branch().frames[current_frame].state`

src/modules/tas_studio/mod.rs
```diff
@@ -1602,6 +1755,12 @@ pub unsafe fn on_tas_playback_frame(
             strafe_state.prev_frame_input.yaw = view_angles[1].to_radians();
         }
 
+        // Store the rendered viewangles
+        // Viewangles is always available. Same as origin.
+        // Vieworigin is something that is not available per player. It is calculated upon
+        // rendering.
```
YaLTeR, line 1761:
```suggestion
        // Store the rendered viewangles.
        // Viewangles are always available. Same as origin.
        // Vieworigin is something that is not available per player. It is calculated upon
        // rendering.
```
Also what do you mean by "same as origin" and also "per player"?

src/hooks/bxt.rs
```diff
@@ -141,6 +141,7 @@ pub struct OnTasPlaybackFrameData {
     pub strafe_cycle_frame_count: u32,
     pub prev_predicted_trace_fractions: [f32; 4],
     pub prev_predicted_trace_normal_zs: [f32; 4],
+    pub rendered_viewangles: [f32; 3],
```
YaLTeR, line 144: Does this need a _v2 function name and a change on bxt side?

src/hooks/engine.rs
khanghugo: I cannot think of how this would matter. The only time both of these are called is when someone loads a campath and then play a campath while they also are playing back TAS views. Talking about stretch.

src/modules/tas_studio/mod.rs
khanghugo: It was fairly subjective because I did make it that way. My reasoning is that since we only change the player's views, none of the stats actually get changed, so none of the HUD info would matter.

src/modules/tas_studio/mod.rs
khanghugo: Playing views doesn't depend on the simulated game or change the TAS in anyway. It receives frames from simulated game then just playback views. I am not very sure what you want to do with playing views states.

src/modules/tas_studio/mod.rs
khanghugo: I don't really know why I wrote these. Deleeted

src/hooks/bxt.rs
khanghugo: Yea we need to change the `on_tas_playback_frame()` function to match. https://github.com/YaLTeR/bxt-rs/pull/101 also makes changes to the playback frame data so this PR (a lot shorter) needs merging first to avoid weird conflicts.

src/modules/tas_studio/mod.rs
YaLTeR: I was thinking that the other game might be doing some simulation while we run playback views, but thinking about it I guess it won't because we already need all accurate frames to play back. Or maybe it can still happen while on a branch?

If that is the reasoning, then replace this blanked match with an explicit PlayingViews to 1. mention in the comment the reasoning, and 2. to not miss if another state is added in the future.

YaLTeR (changes requested):

src/modules/tas_studio/mod.rs
khanghugo: Yea

YaLTeR:
Thanks

PR #100, opened 2024-03-30
Add camera view editing mode

### LINKED ISSUE #96: TAS Editor Camera Editor Camera Editting Mode
Something from TAS editor 1. Should be nice to have it.

khanghugo:
Closes #96 

I think the organizing is a bit weird. Not sure how to go around that because this needs the game itself to work.

Need https://github.com/YaLTeR/BunnymodXT/pull/520 to have it work.

### DIFF
--- a/src/modules/tas_studio/editor/mod.rs
+++ b/src/modules/tas_studio/editor/mod.rs
@@ -309,13 +309,8 @@ struct InsertCameraLineAdjustment {
     did_split: bool,
 }
 
-/// Camera view edit modes.
-///
-/// `CameraViewAdjustmentMode::Alt` adjusts differently for different kinds of line depending on
-/// implementation.
-///
-/// If target_yaw, result is velocity_lock.
-// More to come I guess
+/// `CameraViewAdjustmentMode::Alt` adjusts different things depending on the line type. For
+/// target_yaw lines, it adjusts the velocity_lock angle.
 #[derive(Debug, Clone, Copy)]
 pub enum CameraViewAdjustmentMode {
     Yaw,
@@ -1286,11 +1281,7 @@ impl Editor {
             if let Some(hovered_line_idx) = self.hovered_line_idx {
                 // For camera editor camera line editing.
                 if mouse.buttons.is_right_down() && !self.prev_mouse_state.buttons.is_right_down() {
-                    // Mouse was released last frame so the adjustment cannot be active.
                     assert!(self.camera_view_adjustment.is_none());
-
-                    // There are no other adjustments in the camera editor at the moment; and anyhow
-                    // when more are added, this condition should still be upheld.
                     assert!(!self.is_any_adjustment_active());
 
                     let line = &self.branch().branch.script.lines[hovered_line_idx];
--- a/src/modules/tas_studio/editor/mod.rs
+++ b/src/modules/tas_studio/editor/mod.rs
@@ -26,8 +26,10 @@ use self::utils::{
     bulk_idx_and_repeat_at_frame, join_lines, line_first_frame_idx, line_idx_and_repeat_at_frame,
     FrameBulkExt,
 };
+use super::marker::MainThreadMarker;
 use super::remote::{AccurateFrame, PlayRequest};
-use crate::hooks::sdl::MouseState;
+use crate::hooks::sdl::{self, MouseState};
+use crate::hooks::{bxt, client, engine};
 use crate::modules::tas_optimizer::simulator::Simulator;
 use crate::modules::triangle_drawing::triangle_api::{Primitive, RenderMode};
 use crate::modules::triangle_drawing::TriangleApi;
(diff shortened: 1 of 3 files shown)


src/modules/tas_studio/editor/mod.rs
```diff
@@ -193,8 +202,8 @@ struct ExtraCameraEditorFrameData {
     smoothing_input_region: Option<(usize, usize)>,
     /// Index into `script.lines` of a change frame that ends on this frame.
     change_line_that_ends_here: Vec<usize>,
-    /// Index of a frame where the change line which starts on this frame ends.
-    change_ends_at: Vec<usize>,
+    /// Index of a change line and a frame where the frame is end frame of change line.
+    change_ends_at: Vec<(usize, usize)>,
```
YaLTeR, line 206: Add a separate field?

src/modules/tas_studio/editor/mod.rs
```diff
@@ -1229,6 +1262,75 @@ impl Editor {
                     });
                 }
             }
+
+            if let Some(hovered_line_idx) = self.hovered_line_idx {
+                // For carmera editor camera line editing.
```
YaLTeR, line 1267:
```suggestion
                // For camera editor camera line editing.
```

src/modules/tas_studio/editor/mod.rs
```diff
@@ -1229,6 +1262,75 @@ impl Editor {
...
+                if mouse.buttons.is_right_down() && !self.prev_mouse_state.buttons.is_right_down() {
+                    // Mouse was released last frame so the adjustment cannot be active.
+                    assert!(self.camera_view_adjustment.is_none());
+
+                    // There are no other adjustments in the camera editor at the moment; and anyhow
```
YaLTeR, line 1272: I guess this comment needs an update here and in the other place

src/modules/tas_studio/editor/mod.rs
```diff
@@ -296,6 +305,23 @@ struct InsertCameraLineAdjustment {
     did_split: bool,
 }
 
+/// Camera view edit modes.
+///
+/// `Alt` adjusts differently for different kinds of line.
```
YaLTeR, line 310: Could you describe here what exactly it does?

src/modules/tas_studio/editor/mod.rs
```diff
@@ -4057,6 +4287,33 @@ fn replace_multiple_params<'a>(
...
+
+    view_angles
+}
+
+fn change_origin(origin: Vec3) {
+    let marker = unsafe { MainThreadMarker::new() };
+
+    unsafe { bxt::BXT_TAS_STUDIO_FREECAM_SET_ORIGIN.get(marker)(origin.into()) };
+}
+
```
YaLTeR, line 4316: So, the editor is supposed to be completely independent from the surrounding HL environment. Basically, you should be able to put all Editor code into any other environment and pass the right variables, and it will work.

So for these callbacks, I guess the easiest way is to put them in a struct like
```
struct Callbacks {
    enable_mouse_look: fn(),
    // ...
}
```
and pass this struct into `Editor::tick()`.

src/modules/tas_studio/editor/mod.rs
khanghugo: Alt is in case you want the result to be something like target yaw offset or target yaw velocity lock.

src/modules/tas_studio/editor/mod.rs
khanghugo: What kind of comment change do you prefer here? Straight up deletion?

src/modules/tas_studio/mod.rs
```diff
@@ -2174,3 +2187,25 @@ pub unsafe fn maybe_enable_freecam(marker: MainThreadMarker) {
...
+    sdl::set_relative_mouse_mode(marker, true);
+    client::activate_mouse(marker, true);
+}
+
+fn disable_mouse_look(marker: MainThreadMarker) {
+    sdl::set_relative_mouse_mode(marker, false);
+    client::activate_mouse(marker, false);
+}
+
+fn get_viewangles(marker: MainThreadMarker) -> [f32; 3] {
```
YaLTeR, line 2201: Could you make this and the next function unsafe instead of having the unsafe inside? I think it makes more sense. Then in the callbacks, do `get_viewangles: &|| unsafe { get_viewangles(marker) },`

src/modules/tas_studio/editor/mod.rs
```diff
@@ -296,6 +309,27 @@ struct InsertCameraLineAdjustment {
     did_split: bool,
 }
 
+/// Camera view edit modes.
+///
+/// `CameraViewAdjustmentMode::Alt` adjusts differently for different kinds of line depending on
+/// implementation.
+///
+/// If target_yaw, result is velocity_lock.
+// More to come I guess
```
YaLTeR, line 318:
```suggestion
/// `CameraViewAdjustmentMode::Alt` adjusts different things depending on the line type. For target_yaw lines, it adjusts the velocity_lock angle.
```

src/modules/tas_studio/editor/mod.rs
YaLTeR: I guess delete

YaLTeR:
Thanks

PR #101, opened 2024-04-01
Add Max Acceleration Yaw Offset

khanghugo:
Needs hltas update.

There are some things about this. First, I want that when a frame bulk of this type is split, it shouldn't reset the acceleration values. That is why the weird gymnastics is here. 

Second, I want it a bit easier to use so there's text popping up like #100 . I think it is pretty okay. I don't think the console command for this would be that necessary. 

No third point but I don't want to pass "accurate" data, specifically the current accumulated yawspeed, from BXT to bxt-rs because I think bxt-strafe won't ever mess it up.

I play around with it for a bit and it is very confusing to use. Might get some time getting used to because you have two parameters. I put some thoughts into the interactions and I think it works great for me. Not sure about others.

### DIFF
--- a/bxt-strafe/src/steps.rs
+++ b/bxt-strafe/src/steps.rs
@@ -350,7 +350,7 @@ impl<S: Step> Step for Strafe<S> {
                 frame_bulk.auto_actions.movement
             {
                 let theta = match type_ {
-                    StrafeType::MaxAccel => match dir {
+                    StrafeType::MaxAccel | StrafeType::MaxAccelYawOffset { .. } => match dir {
                         StrafeDir::Left => max_accel_theta(parameters, &state),
                         StrafeDir::Right => -max_accel_theta(parameters, &state),
                         StrafeDir::Yaw(yaw) => {
@@ -445,6 +445,18 @@ impl<S: Step> Step for Strafe<S> {
                     let entry = Vct::get().find_best((vel_yaw + theta) - camera_yaw);
 
                     (camera_yaw, entry)
+                };
+
+                let camera_yaw = if matches!(type_, StrafeType::MaxAccelYawOffset { .. }) {
+                    // theta < 0. = is right
+                    // If is right then we decreases yaw by offset.
+                    // Therefore, positive offset in framebulk mean going more on that side.
+                    let offset = state.max_accel_yaw_offset_value.to_radians();
+                    let offset = if theta < 0. { -offset } else { offset };
+
+                    camera_yaw + angle_mod_rad(offset)
+                } else {
+                    camera_yaw
                 };
 
                 input.yaw = camera_yaw;
@@ -543,6 +555,47 @@ impl<S: Step> Step for ResetFields<S> {
             }))
         ) {
             state.strafe_cycle_frame_count = 0;
+        }
+
+        // If we have some acceleration, then this kicks in.
+        // It will preserve the final value across split segments.
+        if let Some(AutoMovement::Strafe(StrafeSettings {
+            type_:
+                StrafeType::MaxAccelYawOffset {
+                    start,
+                    target,
+                    accel,
+                },
+            dir,
+        })) = frame_bulk.auto_actions.movement
+        {
+            let right = matches!(dir, StrafeDir::Right);
+
+            // Flip start and target when accel is negative.
+            state.max_accel_yaw_offset_value = (state.max_accel_yaw_offset_value + accel)
+                .max(start)
+                .min(target);
+
+            // Reset value if we have different inputs.
+            // This means that if we split a s5x bulk,
+            // there won't be any side effects.
+            if start != state.prev_max_accel_yaw_offset_start
+                || target != state.prev_max_accel_yaw_offset_target
+                || accel != state.prev_max_accel_yaw_offset_accel
+                || right != state.prev_max_accel_yaw_offset_right
+            {
+                state.max_accel_yaw_offset_value = if accel.is_sign_negative() {
+                    target
+                } else {
+                    start
+                };
+
+                // Update so next time we know what to compare against.
+                state.prev_max_accel_yaw_offset_start = start;
+                state.prev_max_accel_yaw_offset_target = target;
+                state.prev_max_accel_yaw_offset_accel = accel;
+                state.prev_max_accel_yaw_offset_right = right;
+            };
         }
 
         self.0
--- a/src/modules/tas_studio/editor/mod.rs
+++ b/src/modules/tas_studio/editor/mod.rs
@@ -9,7 +9,7 @@ use std::time::Instant;
 use bxt_ipc_types::Frame;
 use bxt_strafe::{Hull, Trace};
 use color_eyre::eyre::{self, ensure};
-use glam::{Vec2, Vec3};
+use glam::{IVec2, Vec2, Vec3};
 use hltas::types::{
     AutoMovement, Change, ChangeTarget, Line, StrafeDir, StrafeSettings, StrafeType,
     VectorialStrafingConstraints,
@@ -24,11 +24,12 @@ use self::toggle_auto_action::ToggleAutoActionTarget;
 use self::utils::{
     bulk_and_first_frame_idx, bulk_and_first_frame_idx_mut, bulk_idx_and_is_last,
     bulk_idx_and_repeat_at_frame, join_lines, line_first_frame_idx, line_idx_and_repeat_at_frame,
-    FrameBulkExt,
+    FrameBulkExt, MaxAccelOffsetValuesMut,
 };
 use super::remote::{AccurateFrame, PlayRequest};
 use crate::hooks::sdl::MouseState;
 use crate::modules::tas_optimizer::simulator::Simulator;
+use crate::modules::tas_studio::editor::utils::MaxAccelOffsetValues;
 use crate::modules::triangle_drawing::triangle_api::{Primitive, RenderMode};
 use crate::modules::triangle_drawing::TriangleApi;
 
@@ -127,6 +128,8 @@ pub struct Editor {
     ///
     /// Adjusts the yawspeed in the same way for all adjacent frame bulks with equal yawspeed.
     adjacent_side_strafe_yawspeed_adjustment: Option<AdjacentYawspeedAdjustment>,
+    /// Frame bulk max acceleration yaw offset adjustment.
+    max_accel_yaw_offset_adjustment: Option<MaxAccelYawOffsetAdjustment>,
 
     // ==============================================
     // Camera-editor-specific state.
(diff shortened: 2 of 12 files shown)


khanghugo:
I think this state should be good.

If there's anything I would like to add that would be knowing the yaw speed of previous frame so the current s4x or s5x can continue on that number. Maybe add it now. Maybe add it later. I guess we'd never know.

src/modules/tas_studio/editor/operation.rs
```diff
@@ -12,6 +12,7 @@ use crate::modules::tas_studio::editor::utils::{
 
 // This enum is stored in a SQLite DB as bincode bytes. All changes MUST BE BACKWARDS COMPATIBLE to
 // be able to load old projects.
+// Make sure that newer operations must be added at the end of the enum.
```
YaLTeR, line 15:
```suggestion
// Make sure that newer operations are added at the end of the enum.
```

src/modules/tas_studio/editor/operation.rs
```diff
@@ -102,6 +103,26 @@ pub enum Operation {
         from: Option<String>,
         to: Option<String>,
     },
+    SetMaxAccelStart {
```
YaLTeR, line 106: What about something like
```suggestion
    SetMaxAccelOffsetStart {
```
and others similarly

src/modules/tas_studio/editor/utils.rs
```diff
@@ -24,6 +24,22 @@ pub trait FrameBulkExt {
...
+    // Return a mutable reference to the starting yaw offset, target yaw offset, acceleration,
+    // and original yaw field value stored in the framebulk, if any.
+    fn max_accel_yaw_offset_mut(
+        &mut self,
+    ) -> Option<(
+        &mut f32,
+        &mut f32,
+        &mut f32,
+        Option<&mut f32>,
+        Option<&mut NonZeroU32>,
```
YaLTeR, line 41: Yeah I think at this point it warrants a struct, something like
```
struct MaxAccelOffsetValues<'a> {
    starting_offset: &'a f32,
    // ...
}

struct MaxAccelOffsetValuesMut<'a> {
    starting_offset: &'a mut f32,
    // ...
}
```

src/modules/tas_studio/mod.rs
```diff
@@ -1049,6 +1050,8 @@ Values that you can toggle:
 - s07: right-left strafing
 - s40: constant turn rate to the left
 - s41: constant turn rate to the right
+- s50: accelerated turn rate to the left
+- s51: accelerated turn rate to the right
```
YaLTeR, line 1054:
```suggestion
- s50: accelerated turn to the left
- s51: accelerated turn to the right
```

src/modules/tas_studio/mod.rs
```diff
@@ -1870,6 +1887,13 @@ fn add_frame_bulk_hud_lines(text: &mut Vec<u8>, bulk: &FrameBulk) {
                 StrafeType::ConstYawspeed(yawspeed) => {
                     write!(text, "turn rate: {yawspeed:.0}").unwrap();
                 }
+                StrafeType::MaxAccelYawOffset(start, target, accel) => {
+                    // The values are so tiny that only this would make it sensible.
+                    let start = start * 100.;
+                    let target = target * 100.;
+                    let accel = accel * 100.;
+                    write!(text, "{start:.0} {target:.0} {accel:.2}").unwrap();
```
YaLTeR, line 1895:
```suggestion
                    write!(text, "from: {start:.0}, to: {target:.0}, accel: {accel:.2}").unwrap();
```

src/modules/tas_studio/mod.rs
```diff
@@ -1870,6 +1887,13 @@ fn add_frame_bulk_hud_lines(text: &mut Vec<u8>, bulk: &FrameBulk) {
                 StrafeType::ConstYawspeed(yawspeed) => {
                     write!(text, "turn rate: {yawspeed:.0}").unwrap();
                 }
+                StrafeType::MaxAccelYawOffset(start, target, accel) => {
+                    // The values are so tiny that only this would make it sensible.
```
YaLTeR, line 1891: This is better than just showing them with 2 digits? It could be confusing I guess

src/modules/tas_studio/editor/mod.rs
```diff
@@ -285,6 +287,67 @@ struct AdjacentYawspeedAdjustment {
...
+/// Modes of operation for maximum acceleration yaw offset adjustment.
+///
+/// Cycling by holding right mouse button and left click.
+#[derive(Debug, Clone, Copy)]
+pub enum MaxAccelYawOffsetMode {
+    StartAndTarget = 0,
+    Target,
+    Acceleration,
+    Start,
+    /// Alt is for not left or right strafing.
```
YaLTeR, line 320:
```suggestion
    /// Alt is for adjusting the yaw field value of the strafe direction.
```

src/modules/tas_studio/editor/mod.rs
```diff
@@ -285,6 +287,67 @@ struct AdjacentYawspeedAdjustment {
...
+#[derive(Debug, Clone, Copy)]
+pub struct MaxAccelYawOffsetMouseAdjustment {
+    pub start: f32,
+    pub target: f32,
+    pub accel: f32,
+    pub alt: MaxAccelYawoffsetAlt,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum MaxAccelYawoffsetAlt {
```
YaLTeR, line 333: Maybe like this?
```suggestion
pub enum MaxAccelYawOffsetYawField {
```

src/modules/tas_studio/editor/mod.rs
```diff
@@ -934,22 +1006,93 @@ impl Editor {
...
                             // Presumably, one of the points is invisible, so just fall back.
                             _ => Vec2::X,
                         };
 
-                        if let Some(yaw) = bulk.yaw() {
-                            self.yaw_adjustment = Some(MouseAdjustment::new(*yaw, mouse_pos, dir));
+                        if let Some((start, target, accel)) = bulk.max_accel_yaw_offset() {
+                            // Mode to know which to switch to.
+                            // For left or right, it would be TargetAndEnd so we can turn better.
+                            // For left-right and yaw, it would bet Alt where we can change the
```
YaLTeR, line 1018:
```suggestion
                            // For left-right and yaw, it would be Alt where we can change the
```

src/modules/tas_studio/editor/mod.rs
```diff
@@ -934,22 +1006,93 @@ impl Editor {
...
+                                            };
+
+                                        (
+                                            MaxAccelYawOffsetMode::Alt,
+                                            adjustment_dir,
+                                            MaxAccelYawoffsetAlt::LeftRight(count),
+                                        )
+                                    }
+                                    _ => return Err(ManualOpError::UserError(
+                                        "Max accel yaw offset does not support current strafe dir."
```
YaLTeR, line 1061: Obsolete?

YaLTeR (commented):
Damn, that's a lot of delicate work to make editing for this

src/modules/tas_studio/mod.rs
khanghugo: I tried not having any text for this because it overflows the opaque background. I trust that anyone using this would understand what those numbers mean.

src/modules/tas_studio/editor/mod.rs
khanghugo: Point strafing isn't supported for editor operations because that would adds some more complexity (and effort). I think the message is misinformed. I will change that.

src/modules/tas_studio/mod.rs
khanghugo: Here is a sample part where I use it.

```
s51-------|------|------|0.010000001|- -5.4011025 -5.4011025 0|-|43
s51-------|------|------|0.010000001|- -5.635861 -5.635861 0|-|268
s50-j-----|------|------|0.010000001|- -0.70690674 -0.70690674 0|-|134
s50-j-----|------|------|0.010000001|- -2.2149327 -2.597555 0|-|19
s51-j-----|------|------|0.010000001|- -0.59532917 -0.59532917 0|-|169
s50-j-----|------|------|0.010000001|- -10.585796 -1.0256441 0.12860943|-|982
s01-j-----|------|------|0.010000001|-|-|221
s50-j-----|------|------|0.010000001|- -9.952005 -0.9520051 0.1|-|941
s50-j-----|------|------|0.010000001|- -0.13668747 -0.13668747 0|-|199
s51-j-----|------|------|0.010000001|- 0.32988316 -0.29589385 0|-|295
s51-j-----|------|------|0.010000001|- 2.9388967 4.460467 0|-|7
s51-j-----|------|------|0.010000001|- 2.9388967 4.460467 0|-|72
```

I think 2 digits (times 10) might not be descriptive enough. And also we have space for 3.

bxt-strafe/src/steps.rs
```diff
@@ -545,6 +557,47 @@ impl<S: Step> Step for ResetFields<S> {
...
+                },
+            dir,
+        })) = frame_bulk.auto_actions.movement
+        {
+            let right = matches!(dir, StrafeDir::Right);
+
+            // Flip start and target when accel is negative.
+            state.max_accel_yaw_offset_value = (state.max_accel_yaw_offset_value + accel)
+                .max(start)
+                .min(target);
```
YaLTeR, line 577:
```suggestion
                .clamp(start, target);
```
This should work I think.

src/modules/tas_studio/editor/utils.rs
```diff
@@ -24,6 +24,31 @@ pub trait FrameBulkExt {
...
+    /// Return a reference to the starting yaw offset, target yaw offset, and acceleration stored in
+    /// the framebulk, if any.
+    fn max_accel_yaw_offset(&self) -> Option<MaxAccelOffsetValues>;
+
+    // Return a mutable reference to the starting yaw offset, target yaw offset, acceleration,
+    // and original yaw field value stored in the framebulk, if any.
+    fn max_accel_yaw_offset_mut(&mut self) -> Option<MaxAccelOffsetValuesMut>;
+}
+
+#[allow(dead_code)]
```
YaLTeR, line 37:
```suggestion
```
I think it's used?

src/modules/tas_studio/editor/mod.rs
```diff
@@ -285,6 +288,67 @@ struct AdjacentYawspeedAdjustment {
...
+///
+/// Cycling by holding right mouse button and left click.
+#[derive(Debug, Clone, Copy)]
+pub enum MaxAccelYawOffsetMode {
+    StartAndTarget = 0,
+    Target,
+    Acceleration,
+    Start,
+    /// Alt is for adjusting the yaw field value of the strafe direction.
+    Alt,
```
YaLTeR, line 322: Should this maybe be renamed to YawField or StrafeDirYawField? And in the UI text, something like "Strafe Dir Yaw Field" too?

src/modules/tas_studio/mod.rs
YaLTeR: Maybe write them on 3 lines? This will also allow not multiplying by 100. I'd rather this be understandable without looking at the code

src/modules/tas_studio/editor/utils.rs
khanghugo: `yaw` and `count` members are not used. But they are used in the mutable struct.

YaLTeR: Could you put this on those members?

bxt-strafe/src/steps.rs
khanghugo: This would crash because 

```
Message:  min > max, or either was NaN. min = -0.9614217, max = -1.9672347
```

src/modules/tas_studio/mod.rs
khanghugo: Hmm, I can leave it 1 line with 4 decimal numbers. It looks ok in worst case.

image

khanghugo:
Also waiting for this BXT change https://github.com/YaLTeR/BunnymodXT/pull/519#discussion_r1667319126

I added optimizer support a while ago but I forgot to push so it is now here.

bxt-strafe/src/steps.rs
YaLTeR: Ah, right

YaLTeR (approved):

khanghugo:
Might need https://github.com/YaLTeR/bxt-rs/pull/99 merged first because this PR and https://github.com/YaLTeR/bxt-rs/pull/99 both make changes to BXT exported function.

khanghugo:
Need hltas update and that's it

YaLTeR:
Thanks

PR #110, opened 2024-11-17
TAS studio split and join sections of the script

Eddio0141:
## How to mark a script split
- `save` line
- `reset` line
  - For the sake of smoother editing, once either clients reach this point and reset, it would store the non-shared RNG state to be used on split
- Adding a comment on the script `bxt-rs-split`
  - Difference with the `save` line case is this wouldn't load a save if you are playing the script from before this split marker

## When does the script split
- `bxt_tas_studio_hide` would split the script, if you hide up to the line with the split marker
- You also need a save to be created on either clients

## What happens with bridged hltas
- Nothing, you will always see the full script in this
- Editing the script before the currently used split would invalidate that split save, and rolls back `bxt_tas_studio_hide` to the previous split marker

# Shared RNG
If using a comment split marker, shared RNG is updated to not cause desyncs from shared RNG mismatching after split load 

## Some additional notes
Not doing in this pr, but there has to be a change (or add) to `bxt_tas_studio_hide` to accept value to unhide a number of frames. This is for stopping you from unhiding the entirety of a massive script if you want to go back to some split save

### DIFF
--- a/src/modules/tas_studio/editor/db.rs
+++ b/src/modules/tas_studio/editor/db.rs
@@ -21,6 +21,8 @@ pub struct Branch {
     pub is_hidden: bool,
 
     pub script: HLTAS,
+    pub split_idx: Option<usize>,
+    pub full_script: HLTAS,
     pub stop_frame: u32,
 }
 
@@ -190,11 +192,15 @@ impl Db {
         let script = HLTAS::from_str(&buffer)
             .map_err(|err| eyre!("invalid script value, cannot parse: {err:?}"))?;
 
+        let full_script = script.clone();
+
         Ok(Branch {
             branch_id,
             name,
             is_hidden,
             script,
+            full_script,
+            split_idx: None,
             stop_frame,
         })
     }
@@ -220,12 +226,16 @@ impl Db {
             let script = HLTAS::from_str(&buffer)
                 .map_err(|err| eyre!("invalid script value, cannot parse: {err:?}"))?;
 
+            let full_script = script.clone();
+
             branches.push(Branch {
                 branch_id,
                 name,
                 is_hidden,
                 script,
                 stop_frame,
+                full_script,
+                split_idx: None,
             })
         }
         stmt.finalize()?;
--- a/src/modules/tas_studio/editor/mod.rs
+++ b/src/modules/tas_studio/editor/mod.rs
@@ -11,7 +11,7 @@ use bxt_strafe::{Hull, Trace};
 use color_eyre::eyre::{self, ensure};
 use glam::{IVec2, Vec2, Vec3};
 use hltas::types::{
-    AutoMovement, Change, ChangeTarget, Line, StrafeDir, StrafeSettings, StrafeType,
+    AutoMovement, Change, ChangeTarget, FrameBulk, Line, StrafeDir, StrafeSettings, StrafeType,
     VectorialStrafingConstraints,
 };
 use hltas::HLTAS;
@@ -3460,6 +3460,7 @@ impl Editor {
             return Err(ManualOpError::CannotDoDuringAdjustment);
         }
 
+        // TODO: apply split modifications
         let script = self.script();
         if new_script == *script {
             return Ok(());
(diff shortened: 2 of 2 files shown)


YaLTeR:
I guess there needs to be some command to split while inside the editor.

> bxt_tas_studio_hide would split the script, if you hide up to the line with the split marker

I already forgot if hide does anything other than visually hide, but what exactly happens when you do this and why is it tied to hide? I guess then the editor starts showing only the post-split part?

I kind of imagine splitting to be transparent in the editor, like you put a split marker and it just renders there, then if you edit after the split marker it replays the post-split part, whereas if you edit before, it replays and invalidates the whole thing. Hiding up to the split marker can be just a quality of life command, but why should hiding itself do something?

YaLTeR:
Hmm, yeah, but also you may want some part of the script to show up to make sure you lined it up properly

Eddio0141:
One idea I have is to show an additional set amount of frames from after the bookmark segment ends, maybe something like 250 frames by default and let you customise as you need if you want to peek more into the script, since there could be something like a cutscene you want to wait, or the default is too little because of the low frametime, etc

YaLTeR:
Fwiw I had the same conundrum with seeing too much into the past of the script and decided to only have manual `hide` because you can't really tell what's a good value for all scripts

Eddio0141:
In that case, it could be a command to unhide past the current segment as needed
Something like `bxt_tas_studio_bookmark_unhide 1000` to see 1000 frames after
Avoids showing too much script from choosing segment in the first place, and you could just bind it to repeat
I don't think manual works well in this scenario

src/modules/tas_studio/editor/db.rs
```diff
@@ -190,11 +388,18 @@ impl Db {
         let script = HLTAS::from_str(&buffer)
             .map_err(|err| eyre!("invalid script value, cannot parse: {err:?}"))?;
 
+        let mut splits = SplitInfo::split_lines(script.lines.iter());
+        // TODO: is this fine? not sure how else to get MainThreadMarker for game directory
```
Eddio0141, line 392: Is this fine?

src/modules/tas_studio/editor/db.rs
```diff
@@ -190,11 +388,18 @@ impl Db {
...
+        unsafe {
+            SplitInfo::validate_all_by_saves(&mut splits, MainThreadMarker::new());
+        }
```
Eddio0141, line 395: Is this fine?

src/modules/tas_studio/editor/db.rs
```diff
@@ -35,6 +47,278 @@ impl fmt::Debug for Branch {
...
+    // for the sake of easier searching, the last framebulk index is stored
+    pub bulk_idx: usize,
+    // TODO: could probably get away with &str
+    pub name: String,
+    pub split_type: SplitType,
+    // ready as in, there's a save created, and lines before and including start_idx is still unchanged
+    pub ready: bool,
+    // TODO: timing on grabbing rng values
+    // TODO: how do i get shared rng from engine?
+    // TODO: i most likely need to add this information being sent from a sim client, how can this be done in a clean way
```
Eddio0141, line 66: I check `SplitInfo` for ready state inside `apply_accurate_frame`, I need shared and non-shared RNG so splits can use those information
Is it fine to just give those information in `AccurateFrame` struct? I don't know the alternate way to grab those information from whatever is sending `AccurateFrame`

src/modules/tas_studio/editor/db.rs
```diff
@@ -35,6 +47,278 @@ impl fmt::Debug for Branch {
...
+    pub shared_rng: Option<u32>,
```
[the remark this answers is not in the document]
khanghugo: Shared seed seems to be in server DLL. Looking at BXT, it hooks into CmdStart and manipulate the seed there.

You can get the seed directly from glSeed. I have a BXT pr that deals with manipulating random stuffs
https://github.com/YaLTeR/BunnymodXT/pull/418/

You can hook some client and server stuffs just fine with bxt-rs. Right now it just does not read from client or server libraries so you have to write some of that code yourself. I did do that before but iirc I scrapped my work so I won't have any starter code for you.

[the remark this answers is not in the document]
Eddio0141: Nice, thanks. `CmdStart` seems to already be hooked in bxt-rs, I might as well as just use that

PR #112, opened 2024-11-26
Fix colour values not being decoded/encoded

GameChaos:
Colloquially known as "gamma correction".

Apologies if my code is poopy or organised in the wrong place.

Colour values need to be decoded with the correct transfer function before blending or doing other stuff to them (and later encoded with the [OETF (opto electrical transfer function)](https://en.wikipedia.org/wiki/Transfer_functions_in_imaging) for display on a monitor).

The sRGB standard defines piecewise and pure power curves as transfer functions. I've opted to use the power curves, because most sRGB monitors use the pure power curve as an EOTF (x^(1/2.2)). More info can be found here with people arguing if it's right or not :) https://community.acescentral.com/t/srgb-piece-wise-eotf-vs-pure-gamma/4024

Here are some images for comparison (screenshots from recorded video, so quality is a bit meh):
image
image

The first one is where no encoding/decoding is done, so things will look darker and "energy" isn't preserved.


khanghugo:
+1. Sampling recording does look a lot darker than non sampling and I think this has been neglected for a while.

GameChaos:
> I assume this is a similar fix to [crashfort/SourceDemoRender#68](https://github.com/crashfort/SourceDemoRender/issues/68)?

yes, same fix

comparison videos:
https://www.youtube.com/watch?v=Tl01yt_Uud4
https://www.youtube.com/watch?v=TaTIonPFT5k

looked at the cpu fallback code, seems way too complex for me to handle :/

YaLTeR:
Could you upload .mp4 here on github? Without YT compression

GameChaos:
> Could you upload .mp4 here on github? Without YT compression

files are too big :(

put them on mediafire: https://www.mediafire.com/file/g4fxqhmdgbp4i6m/blending+clips.zip/file

YaLTeR:
As I understand, stationary frames should not change color, but they do, for example red on the very first frame:

image

GameChaos:
oh crap. could it be because of quantisation? i'll take a look soon :tm:

PR #116, opened 2025-03-01
fix bindgen

khanghugo:
Bindgen 0.71 and latest rust nightly version.

Bindgen does not right away so there are some changes to bindgen such as `c_uint` to `c_int` or `*mut T` to `*const T`. With that, bxt-rs code can remain unchanged. Bindgen also makes function type to be `Option<unsafe extern ...>` instead of the usual `unsafe extern ...`. So, for some parts, I add `.unwrap()` wherever applicable. There are a lot of changes needed doing after generating bindings so I write all instructions along with the bindgen command inside each binding files.

Some files are deleted because the new bindgen conveniently generates bunch of related structs. There are some re-declared structs such as `hull_s` where its members are defined in some files and not defined in some files. Doesn't seem like a problem.

I am able to load a TAS and interact with the editor just fine. If anything to worry about, it is the `tas_server_time_fix` because there is a small change to the code that does pointer comparison. Though the changed code is from rust_analyzer suggestion so I guess it is all okay.


YaLTeR:
Oh right, I forgot I did a bunch of manual cleanup on the generated bindings. Thanks for documenting the steps.

> Bindgen also makes function type to be Option<unsafe extern ...> instead of the usual unsafe extern .... So, for some parts, I add .unwrap() wherever applicable.

I got rid of all `Option<>` manually in the generated bindings, I think that's a better move here rather than adding unwrap(), since we know that those pointers are always present. (Only leave `Option<>` when the pointer can actually be NULL.)

I also replaced all `::std::os::raw::` and such with `use` manually, it's easy enough, just do a search and replace for `::std::os::raw::` to nothing, and same for `::std::mem::` (add the new `offset_of` import). `::std::option::` can also be removed.

For types like `vec3_t` I replaced them with `[f32; 3]` manually to avoid these spammy type aliases. Also doable with a search and replace. Same with edict_s -> edict_t etc.

Remove `Copy` and `Clone` too since the types are not semantically copyable.

Also if you don't mind, move structs like `usercmd_s` back to their separate files.

With these cleanups it should be easy to see if anything changed in the bindings unexpectedly. Actually I don't think much should change in the actual types, it's the layout tests that changed.

I don't think we need to regen this often, so it's fine that more manual work is involved. Actually I doubt we'll have to regen this for a while if at all because the major problem with bindgen layout tests had been fixed now with `offset_of!` and I don't think there are any other big problems.

khanghugo:
Take a look again and see. Doesn't seem to have any problem loading and interacting with a TAS

src/ffi/com_model.rs
```diff
@@ -846,204 +252,44 @@ pub struct msurface_s {
...
+    ["Offset of field: msurface_s::texturemins"][offset_of!(msurface_s, texturemins) - 44usize];
+    ["Offset of field: msurface_s::extents"][offset_of!(msurface_s, extents) - 48usize];
+    ["Offset of field: msurface_s::texinfo"][offset_of!(msurface_s, texinfo) - 52usize];
+    ["Offset of field: msurface_s::styles"][offset_of!(msurface_s, styles) - 56usize];
+    ["Offset of field: msurface_s::samples"][offset_of!(msurface_s, samples) - 60usize];
+    ["Offset of field: msurface_s::pdecals"][offset_of!(msurface_s, pdecals) - 64usize];
+};
 #[repr(C)]
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug)]
```
YaLTeR, line 279:
```suggestion
#[derive(Debug, Copy, Clone)]
```
Can remain Clone/Copy because no pointers

src/ffi/playermove.rs
```diff
@@ -28,11 +49,61 @@ pub const PM_TRACELINE_PHYSENTONLY: TraceLineFlag = 0;
...
+/* automatically generated by rust-bindgen 0.71.1 */
+
+#[doc = "\tCopyright (c) 1996-2002, Valve LLC. All rights reserved.\n\n\tThis product contains software technology licensed from Id\n\tSoftware, Inc. (\"Id Technology\").  Id Technology (c) 1996 Id Software, Inc.\n\tAll Rights Reserved.\n\n   Use, distribution, and modification of this source code and/or resulting\n   object code is restricted to non-commercial enhancements to products from\n   Valve LLC.  All other use, distribution, or modification is prohibited\n   without written permission from Valve LLC."]
+pub type string_t = c_uint;
+pub type byte = c_uchar;
+pub const qboolean_false_: qboolean = 0;
+pub const qboolean_true_: qboolean = 1;
+pub type qboolean = c_uint;
+#[repr(C)]
+#[derive(Debug)]
```
YaLTeR, line 61:
```suggestion
#[derive(Debug, Copy, Clone)]
```
Can remain Clone/Copy because no pointers

src/ffi/playermove.rs
```diff
@@ -110,46 +181,43 @@ pub struct playermove_s {
...
         traceFlags: TraceFlags,
         ignore_pe: c_int,
     ) -> pmtrace_s,
     pub PM_TraceLine: unsafe extern "C" fn(
-        start: *const f32,
-        end: *const f32,
-        flags: TraceLineFlag,
+        start: *mut f32,
+        end: *mut f32,
+        flags: c_int,
```
YaLTeR, line 194: I think this should remain as it was?

src/ffi/playermove.rs
```diff
@@ -28,11 +49,61 @@ pub const PM_TRACELINE_PHYSENTONLY: TraceLineFlag = 0;
...
+    ["Offset of field: trace_t::startsolid"][offset_of!(trace_t, startsolid) - 4usize];
+    ["Offset of field: trace_t::inopen"][offset_of!(trace_t, inopen) - 8usize];
+    ["Offset of field: trace_t::inwater"][offset_of!(trace_t, inwater) - 12usize];
+    ["Offset of field: trace_t::fraction"][offset_of!(trace_t, fraction) - 16usize];
+    ["Offset of field: trace_t::endpos"][offset_of!(trace_t, endpos) - 20usize];
+    ["Offset of field: trace_t::plane"][offset_of!(trace_t, plane) - 32usize];
+    ["Offset of field: trace_t::ent"][offset_of!(trace_t, ent) - 48usize];
+    ["Offset of field: trace_t::hitgroup"][offset_of!(trace_t, hitgroup) - 52usize];
+};
+pub type int32 = c_int;
```
YaLTeR, line 100: I'd get rid of this, will remove some changes too

src/ffi/playermove.rs
```diff
@@ -175,9 +243,9 @@ pub struct playermove_s {
         bparam2: c_int,
     ),
     pub PM_PlayerTraceEx: unsafe extern "C" fn(
-        start: *const f32,
-        end: *const f32,
-        traceFlags: TraceFlags,
+        start: *mut f32,
+        end: *mut f32,
+        traceFlags: c_int,
```
YaLTeR, line 248: Should also remain as it was

src/ffi/playermove.rs
```diff
@@ -186,995 +254,164 @@ pub struct playermove_s {
         pfnIgnore: unsafe extern "C" fn(pe: *mut physent_s) -> c_int,
     ) -> c_int,
     pub PM_TraceLineEx: unsafe extern "C" fn(
-        start: *const f32,
-        end: *const f32,
-        flags: TraceLineFlag,
+        start: *mut f32,
+        end: *mut f32,
+        flags: c_int,
```
YaLTeR, line 259: Should also remain as it was

src/ffi/pmplane.rs
```diff
@@ -1,46 +1,20 @@
...
+#![allow(unused, nonstandard_style, deref_nullptr)]
+
+use std::mem::{align_of, offset_of, size_of};
 use std::os::raw::*;
-use std::ptr::null;
 
 #[repr(C)]
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug)]
 pub struct pmplane_t {
```
YaLTeR, line 10: Is it the same struct, just with a C typedef or something? If so then can get rid of it

src/ffi/pmtrace.rs
```diff
@@ -1,136 +1,42 @@
...
+use super::pmplane::pmplane_t;
+
+pub const qboolean_false_: qboolean = 0;
+pub const qboolean_true_: qboolean = 1;
+pub type qboolean = c_uint;
 
 #[repr(C)]
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug)]
 pub struct pmtrace_s {
```
YaLTeR, line 16: Same, if it's the same struct but with a C typedef, then get rid of one of these

src/ffi/triangleapi.rs
```diff
@@ -43,227 +58,43 @@ pub struct triangleapi_s {
...
+    pub SpriteTexture: unsafe extern "C" fn(pSpriteModel: *mut model_s, frame: c_int) -> c_int,
     pub WorldToScreen: unsafe extern "C" fn(world: *const f32, screen: *mut f32) -> c_int,
-    pub Fog: unsafe extern "C" fn(flFogColor: *const f32, flStart: f32, flEnd: f32, bOn: c_int),
+    pub Fog: unsafe extern "C" fn(flFogColor: *mut f32, flStart: f32, flEnd: f32, bOn: c_int),
     pub ScreenToWorld: unsafe extern "C" fn(screen: *const f32, world: *mut f32),
     pub GetMatrix: unsafe extern "C" fn(pname: c_int, matrix: *mut f32),
-    pub BoxInPVS: unsafe extern "C" fn(mins: *const f32, maxs: *const f32) -> c_int,
-    pub LightAtPoint: unsafe extern "C" fn(pos: *const f32, value: *mut f32),
+    pub BoxInPVS: unsafe extern "C" fn(mins: *mut f32, maxs: *mut f32) -> c_int,
+    pub LightAtPoint: unsafe extern "C" fn(pos: *mut f32, value: *mut f32),
```
YaLTeR, line 67: I think I manually changed these muts to consts after checking the decompile. So it should be fine to change them to const.

YaLTeR (commented):
Thanks, looks much better. Left some suggestions and comments

src/ffi/playermove.rs
khanghugo: I did not change all of these. It is the bindgen that changed it. But yea, I understand why it is more desireable for it to be const.

src/ffi/pmplane.rs
khanghugo: Not sure what you mean by this. This is the same struct from the older bxt-rs bindgen. Albeit, pmplane.rs and pmtrace.rs aren't referenced anywhere outside of this `ffi` module.

src/ffi/playermove.rs
YaLTeR: Yeah, I believe I changed those by hand to what they are on current main

src/ffi/pmplane.rs
YaLTeR: I mean, where does it come from the includes? Is it actually defined twice in two different C includes? Or one of the includes defines it, and some other include just typedefs it under a slightly different name (pmplane vs. plane)

khanghugo: It seems like `pmplane_t` and `plane_t` do have exactly same data structure, but they are used differently. They are not type alias of one another. `trace_t` and `pmtrace_t` look similar but they have slightly different data structure. I think it is better just to keep things as is.

src/ffi/playermove.rs
```diff
@@ -177,7 +244,7 @@ pub struct playermove_s {
     pub PM_PlayerTraceEx: unsafe extern "C" fn(
         start: *const f32,
         end: *const f32,
-        traceFlags: TraceFlags,
+        traceFlags: c_int,
```
YaLTeR, line 247:
```suggestion
        traceFlags: TraceFlags,
```

src/ffi/playermove.rs
```diff
@@ -188,993 +255,162 @@ pub struct playermove_s {
     pub PM_TraceLineEx: unsafe extern "C" fn(
         start: *const f32,
         end: *const f32,
-        flags: TraceLineFlag,
+        flags: c_int,
```
YaLTeR, line 258:
```suggestion
        flags: TraceLineFlag,
```

src/ffi/playermove.rs
```diff
@@ -120,36 +190,33 @@ pub struct playermove_s {
     pub PM_TraceLine: unsafe extern "C" fn(
         start: *const f32,
         end: *const f32,
-        flags: TraceLineFlag,
+        flags: c_int,
```
YaLTeR, line 193:
```suggestion
        flags: TraceLineFlag,
```

YaLTeR:
Thanks, I hope I didn't break anything with those last changes

PR #120, opened 2025-03-14
add show_player_in_hltv

### LINKED ISSUE #117: Draw third person model in dem_forcehltv freecam
Already implemented in BXT ([here](https://github.com/YaLTeR/BunnymodXT/pull/238)) but it shouldn't be hard to implement directly in bxt-rs. The current implementation is taken from [HLAE](https://github.com/advancedfx/advancedfx/blob/main/AfxHookGoldSrc/HltvFix.cpp) iirc.
This would be useful for recording, and implementing it wouldn't hurt, so people wouldn't need to inject BXT.

khanghugo:
closes #117 

The implementation is very similar to BXT. I also realize that I have been taking this feature for granted, recording a few demos with this feature without knowing it is a feature.

There is `Cmd_FindCmd` but I just feel like it is easier to just implement my own.

### DIFF
--- a/src/hooks/engine.rs
+++ b/src/hooks/engine.rs
@@ -2,7 +2,7 @@
 
 #![allow(non_snake_case, non_upper_case_globals)]
 
-use std::ffi::CString;
+use std::ffi::{CStr, CString};
 use std::fmt;
 use std::num::ParseIntError;
 use std::os::raw::*;
@@ -99,6 +99,20 @@ pub static CL_GameDir_f: Pointer<unsafe extern "C" fn()> = Pointer::empty_patter
     ]),
     null_mut(),
 );
+pub static CL_IsSpectateOnly: Pointer<unsafe extern "C" fn() -> c_int> = Pointer::empty_patterns(
+    b"CL_IsSpectateOnly\0",
+    // To find, search for "Redirected to invalid server".
+    // You are inside CL_SignonReply. This string is printed inside a conditional scope based on
+    // two conditions. The value that is equal compared to a false value is cls.spectator.
+    // Cycle through references of cls.spectator until you find a short function that has exactly
+    // one function call and returns not equal comparison between cls.spectator and a false
+    // value.
+    Patterns(&[
+        // 8684
+        pattern!(FF 15 ?? ?? ?? ?? 8B 0D ?? ?? ?? ?? 33 C0 85 C9 0F 95 C0),
+    ]),
+    null_mut(),
+);
 pub static cl_lightstyle: Pointer<*mut [lightstyle_t; 64]> = Pointer::empty(b"cl_lightstyle\0");
 pub static CL_Move: Pointer<unsafe extern "C" fn()> = Pointer::empty_patterns(
     b"CL_Move\0",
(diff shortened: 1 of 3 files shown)


src/hooks/engine.rs
```diff
@@ -1460,6 +1489,34 @@ pub unsafe fn player_edict(marker: MainThreadMarker) -> Option<NonNull<edict_s>>
...
+        return None;
+    }
+
+    let mut curr_cmd_function = *cmd_functions_;
+
+    loop {
+        let cmd_function = &*curr_cmd_function;
+        let curr_name = CStr::from_ptr(cmd_function.name as *mut i8);
+
+        if curr_name.to_str().ok()? == name {
```
YaLTeR, line 1506: This shouldn't early-return upon to_str() failure, it just means we need to skip this cmd.
```suggestion
        if curr_name.to_str() == Some(name) {
```

src/modules/show_player_in_hltv.rs
```diff
@@ -0,0 +1,47 @@
+//! `Show player model in HLTV spectate`
```
YaLTeR, line 1:
```suggestion
//! Show player model in HLTV spectate.
```

src/modules/show_player_in_hltv.rs
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    fn is_enabled(&self, marker: MainThreadMarker) -> bool {
+        // BXT has the same functionality.
+        !bxt::BXT_TAS_LOAD_SCRIPT_FROM_STRING.is_set(marker)
+            && engine::cls_demos.is_set(marker)
+            && engine::ClientDLL_IsThirdPerson.is_set(marker)
+            && engine::CL_IsSpectateOnly.is_set(marker)
+            && engine::pmove.is_set(marker)
+            && unsafe { engine::find_cmd(marker, "dem_forcehltv") }.is_some()
```
YaLTeR, line 33: Hmm, I don't like this logic involving some engine state in `is_enabled()`. Perhaps should find this somewhere similar to how we find pointers?

src/modules/show_player_in_hltv.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn description(&self) -> &'static str {
+        "\
+When people do cam path, they usually want to see the player in the demo. This means the demo will be recorded in third person. \
+The only way to enter third person and free camera is to do `dem_forcehltv 1` and then use `viewdemo` command. \
+For a brief moment in free spectate in `viewdemo`, the player model will show up. \
+However, after cycling through all spectating options once, the player model disappears.
+
+This is the fix for that.
+
+Alternatively, with this fix, it is possible to use `playdemo` with `thirdperson` to have the player model show up to just record. \
```
YaLTeR, line 22: Does this mean the `dem_forcehltv 1` check is not needed?

src/modules/show_player_in_hltv.rs
khanghugo: These HLTV and demo codes are inside demoplayer.so (or at the very least, not inside engine code). I don't see any references to static HLTV variables inside engine. The reason why there is cvar check in the first place is because BXT does it. I suspect it is to avoid crashing or unintended behaviors. @SmileyAG may know why it is needed.

Anyways, if it doesn't look like what you want, it is possible to store a static variable caching the result of this find_cmd during run time to check that `dem_forcehltv` actually exists.

src/modules/show_player_in_hltv.rs
YaLTeR: Then maybe remove the whole dem_forcehltv check and find_cmd?

khanghugo:
After some more testing on this, it seems like `playdemo` and then `thirdperson` displays the player model without the fix. `viewdemo` and `dem_forcehltv 1` route does need this fix.

The reason why this PR/fix exists is so that people can record a demo in third person. Personally, after learning that `playdemo` and `thirdperson` works just fine, I would just stick with that. However, some people may prefer the HLTV way and this fix is for them.

src/modules/show_player_in_hltv.rs
```diff
@@ -0,0 +1,47 @@
...
+When record a demo with camera path, it is desirable for the demo to be in third person. \
+To enter third person in a demo, there are two options.
+
+`playdemo` and `thirdperson`. If there is campath loaded, this is the most straightforward way to record. \
+The player model will always show up. But just like any `playdemo`, you can only play a demo.
+
+`viewdemo` and `dem_forcehltv 1`. There are more functionalities available with this method such as pausing, fastforwarding, and rewinding. \
+However, the player model does not consistently show up.
+
+This fix is for `viewdemo` and `dem_forcehltv 1` route. This fix forces the player model to show up."
```
YaLTeR, line 24:
```suggestion
When recording a demo with a custom camera path, you frequently want to see the player in third person. \
There are two ways to do this.

- `playdemo` and `thirdperson`. If a campath is loaded, this is the most straightforward way to record. \
The player model will always show up.

- `viewdemo` and `dem_forcehltv 1`. This gives you more functionality such as pausing, fast-forwarding, and rewinding. \
However, the player model is not consistently visible.

This module makes the player model consistently visible in `viewdemo` with `dem_forcehltv 1`."
```

src/modules/show_player_in_hltv.rs
```diff
@@ -0,0 +1,47 @@
+//! Show player model in HLTV spectate
```
YaLTeR, line 1:
```suggestion
//! Show player model in HLTV spectate.
```

YaLTeR:
If I wrote ok, then commit via github UI and let's merge

khanghugo:
Let's do it.

YaLTeR:
Thanks

PR #122, opened 2025-05-19
campath: add option to rotate camera points around origin Z axis

khanghugo:
To create a campath file inside Blender, there must be a map file in Blender. Usually, people use "Nem's Crafty" to export GoldSrc BSP into OBJ. However, Crafty is for Source. GoldSrc 90 degrees yaw is Source 0 degrees yaw or something. 

If people were to have the source file of the map, which is JMF or MAP, and they export that inside Blender, the OBJ has correct orientation for GoldSrc. And when exporting a campath created based on that source map file, HLAE Blender Plugins can mess up the rotation. The result is the campath and rotation look right in Blender, but not in the game.

After rotating the points with basic math, need to change the viewangles as well.

### DIFF
--- a/src/modules/campath/mod.rs
+++ b/src/modules/campath/mod.rs
@@ -38,7 +38,7 @@ impl Module for Campath {
     }
 
     fn cvars(&self) -> &'static [&'static CVar] {
-        static CVARS: &[&CVar] = &[&BXT_CAMPATH_OFFSET, &BXT_CAMPATH_LOAD];
+        static CVARS: &[&CVar] = &[&BXT_CAMPATH_OFFSET, &BXT_CAMPATH_LOAD, &BXT_CAMPATH_ROTATE];
         CVARS
     }
 
@@ -69,6 +69,18 @@ static BXT_CAMPATH_OFFSET: CVar = CVar::new(
     b"-0.25\0",
     "\
 Offsets time values in second of each campath entry. Offset can be negative. Default is -0.25.
+",
+);
+
+static BXT_CAMPATH_ROTATE: CVar = CVar::new(
+    b"bxt_campath_rotate\0",
+    b"0\0",
+    "\
+Rotates all camera points around origin by Z up axis. HLAE CAM format is mainly for Source. But, Source horizontal rotation is slightly different from GoldSrc.
+
+If you have created campath in Blender based on map file exported from TrenchBroom or jack, you should set this value to 90. 
+
+Other tools such as Nem's Crafty are meant for Source. Its .OBJ export implicitly adds rotation. 
 ",
 );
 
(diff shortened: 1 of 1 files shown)


src/modules/campath/mod.rs
```diff
@@ -162,6 +176,17 @@ fn force_live(marker: MainThreadMarker) {
     IS_FORCED.set(marker, true);
 }
 
+fn rotate_round_z(point: glam::Vec3, rad: f32) -> glam::Vec3 {
+    // only change x and y
+    let orig_x = point.x;
+    let orig_y = point.y;
+
+    let x = orig_x * rad.cos() - orig_y * rad.sin();
```
YaLTeR, line 184: Could probably use https://docs.rs/glam/latest/glam/f32/struct.Mat2.html#method.from_angle and then multiply by it

src/modules/campath/mod.rs
```diff
@@ -162,6 +176,17 @@ fn force_live(marker: MainThreadMarker) {
     IS_FORCED.set(marker, true);
 }
 
+fn rotate_round_z(point: glam::Vec3, rad: f32) -> glam::Vec3 {
```
YaLTeR, line 179: rotate_about_z?

src/modules/campath/mod.rs
```diff
@@ -72,6 +72,20 @@ Offsets time values in second of each campath entry. Offset can be negative. Def
 ",
 );
 
+static BXT_CAMPATH_ROTATE: CVar = CVar::new(
+    b"bxt_campath_rotate\0",
+    b"0\0",
+    "\
+Rotates all camera points around origin by Z up axis. HLAE CAM format is mainly for Source. But, Source horizontal rotation is slightly different from GoldSrc.
```
YaLTeR, line 79:
```suggestion
Rotates all camera points around origin by Z up axis.

HLAE CAM format is mainly for Source. But, Source horizontal rotation is slightly different from GoldSrc.
```

I'm also not sure what "Rotates all camera points around origin by Z up axis." means. Shouldn't this say something like "Rotate all camera points about (some specific axis) by this angle."?

khanghugo: It means exactly what you think. I just can't quite unwrap the technical details so people may understand why they need to do this.

src/modules/campath/mod.rs
khanghugo: it is supposed to be "rotate_around_z" but I think I forget an "a" there.

src/modules/campath/mod.rs
YaLTeR: Maybe "Rotates all camera points about the Z-up axis at the origin by this angle in degrees."?

YaLTeR:
Thanks

PR #133, opened 2026-01-06
Allow launching multiple instances without BunnymodXT

khanghugo:
Make it easier for people to use just bxt-rs. 

One incidental good usage of BunnymodXT is that it always "focuses" on the game when the game is not in focus, meaning no FPS drop. BXT also allows launching multiple instances. With those two combined, a streaming setup containing multiple views of the game.

This should be included in bxt-rs as some of the more features (for spectating) I recently developed for are only on bxt-rs.

### DIFF
--- a/src/hooks/engine.rs
+++ b/src/hooks/engine.rs
@@ -2086,6 +2086,41 @@ fn reset_pointers(marker: MainThreadMarker) {
     }
 }
 
+#[cfg(windows)]
+fn allow_multiple_instances() {
+    // trying to free launcher mutex
+    // to have multiple game instances
+    // BXT also tries doing the same thing but there is no conflict
+
+    use winapi::um::handleapi::CloseHandle;
+    use winapi::um::synchapi::{OpenMutexW, ReleaseMutex};
+    use winapi::um::winnt::SYNCHRONIZE;
+
+    fn to_wide(s: &str) -> Vec<u16> {
+        use std::ffi::OsStr;
+        use std::os::windows::ffi::OsStrExt;
+
+        OsStr::new(s)
+            .encode_wide()
+            // add null terminator
+            .chain(std::iter::once(0))
+            .collect()
+    }
+
+    let mutex_name = to_wide("ValveHalfLifeLauncherMutex");
+
+    // this mutex exists so need to open the exact mutex
+    // with OpenMutexW, not OpenMutexA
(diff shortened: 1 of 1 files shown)


YaLTeR:
Isn't this function called for every single hooked pointer? Doesn't seem appropriate to do this here

khanghugo:
You are absolutely right -- your assertion correctly points out implementation inefficiency from my part!

Anyway, fixed. Interesting that `OpenMutexA()` does not work if freeing mutex happens once outside the pointer hooking function. Though I do know that `OpenMutexW()` exists because it is the first google result when I look up "OpenMutexA". And that works.

src/hooks/engine.rs
```diff
@@ -2118,6 +2118,40 @@ pub mod exported {
...
+                            .chain(std::iter::once(0))
+                            .collect()
+                    }
+
+                    // this mutex exists so need to open the exact mutex
+                    // with OpenMutexW, not OpenMutexA
+                    let mutex = OpenMutexW(
+                        SYNCHRONIZE,
+                        0,
+                        to_wide("ValveHalfLifeLauncherMutex").as_ptr(),
```
YaLTeR: Pretty sure this will drop the Vec before it gets passed into a function, leading to a use after free

Also while we're at it could you extract all this into a separate function

khanghugo: Hmmmmm, I thought compiler would just create a constant string during compilation. rust-analyzer didn't complain and it works. This code is copied from my other project and that also works.

YaLTeR: I mean maybe it could be optimized but here you're allocating a new Vec<> that you're returning from a function, getting its .as_ptr(), then instantly dropping the Vec, and passing the now-stale ptr into the function. If you try that with CString::new() I think they even added a warning/error for this

YaLTeR:
> Interesting that `OpenMutexA()` does not work if freeing mutex happens once outside the pointer hooking function.

But BXT uses A? https://github.com/YaLTeR/BunnymodXT/blob/eec19bf169f77b813724118c9c4b009fdecb4077/BunnymodXT/helper_functions.hpp#L23

khanghugo:
Yes it does. I am still not sure why but when I use `OpenMutexA()` inside the loop, it works. When I use it outside the loop, it no works. `OpenMutexW()` however works in that case. I decide to stop my investigation once it works. 
That is on Linux. I had to willingly boot up Windows for testing and the same behavior exhibits.

YaLTeR:
Weird and unfortunate to have that vector but alright ty

